PR #40, opened 2022-08-14
FIX: when host is https set useSSL as true

sidmohanty11:
- added a regex expression to check whether host is https or HTTP and set use SSL according to that


RonLek:
@sidmohanty11 could you please add a description and resolve merge conflicts.

sidmohanty11:
@RonLek done

Sing-Li:
merging on behalf of @RonLek

PR #46, opened 2022-08-26
FIX: when cookies are there assume user present

sidmohanty11:
- use `js-cookie` instead of an instance (RCInstance) to get the cookies so that if cookies are present then the user is assumed to be logged in using some source


Sing-Li:
Emergency fix for Demo Day.   Merging on behalf of @RonLek

PR #47, opened 2022-08-26
CHORE: remove tech.co.html

sidmohanty11:
- Remove tech.co.html which we will use in the demo repo for publishing


Sing-Li:
Merging for emergency Demo Day fix  on behalf of @RonLek

PR #51, opened 2022-10-01
[FIX] add user data if not present

sidmohanty11:
A `/me` call when user data is not present


RonLek:
@sidmohanty11 PTAL at the conflicts.

PR #84, opened 2023-01-13
Displaying joining or leaving message properly

### LINKED ISSUE #83: Not displaying joining or leaving message properly
### Description:

Whenever a user leave or join a channel or removed by someone from the channel, then the message is not displaying properly.
### Steps to reproduce:

1. Join/Leave the channel through the RC server.
2. Login in embedded chat through any account.
3. See the joining/leaving message.

### Expected behavior:

The message should be something like this:
- In case of joining - `@username joined the channel`
- In case of leaving - `@username left the channel`
- In case of removing by admin/owner - `@adminUsername removed @username`

### Actual behavior:

Only the username is printing on the screen with no clear message either the user has joined the channel or left the channel.

vishakha-aggarwal:
# Displaying joining or leaving messages properly

- [X] Printing joining and removing messages properly

Fixes #83


vishakha-aggarwal:
@sidmohanty11 please review my PR. Thank you.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "embeddedchat",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
sidmohanty11, line 3: I think I made this change already

src/components/MessageList/MessageList.js
```diff
@@ -121,91 +121,155 @@ const MessageList = ({ messages, handleGoBack }) => {
...
-                            )
-                          }
-                        >
-                          <Markdown body={reaction.name} />
-                          <p>{reaction.count}</p>
-                        </MessageReactions.Reaction>
-                      ))}
-                  </MessageReactions>
+                  {msg.t === undefined ? (
+                    <>
```
sidmohanty11, line 125: I don't think we need this fragment

src/components/MessageList/MessageList.js
```diff
@@ -121,91 +121,155 @@ const MessageList = ({ messages, handleGoBack }) => {
...
-                              !isSameUser(reaction, authenticatedUserUsername)
-                            )
-                          }
-                        >
-                          <Markdown body={reaction.name} />
-                          <p>{reaction.count}</p>
-                        </MessageReactions.Reaction>
-                      ))}
-                  </MessageReactions>
+                  {msg.t === undefined ? (
```
sidmohanty11, line 124: Please don't explicitly check for `undefined`, can we do `!msg.t` ?

src/components/MessageList/MessageList.js
```diff
@@ -121,91 +121,155 @@ const MessageList = ({ messages, handleGoBack }) => {
...
+                            </MessageReactions.Reaction>
+                          ))}
+                      </MessageReactions>
+                    </>
+                  ) : (
+                    <>
+                      {msg.t === 'ul' && (
+                        <Message.Header>
+                          <Message.Name>@{msg.u.username} </Message.Name>
+                          <Message.Username style={{ marginLeft: '2px' }}>
```
sidmohanty11, line 174: You can see that all these blocks contain exactly the same code except the username part, so IMO we can just make a function of a sort where we can have a switch case to check what is the condition and according to that return the `<Message.Username>{userActions(msg.t)}</Message.Username>`

Something like,
```
function userActions(type) { // its msg.t, please assign it a better name if you think of any
  switch type {
    case 'uj':
        return 'joined the channel' // something like this
    ...
  }
}
```

What do you think?

sidmohanty11 (changes requested):
Thank you @vishakha-aggarwal for this PR!

src/components/MessageList/MessageList.js
vishakha-aggarwal: Yes, this would make more sense. Thanks!

src/components/MessageList/MessageList.js
vishakha-aggarwal: I think this fragment is necessary, without this fragment we will not be able to wrap up the children components under parent component. Or is there any way to wrap the components without using this fragment? If there is any way, then please elaborate me a little bit so that I could implement that. Thanks!

sidmohanty11: Ahh sorry this is not the fragment that I was talking about,


Line 170
```
   <>
                      {msg.t === 'ul' && (
                        <Message.Header>
                          <Message.Name>@{msg.u.username} </Message.Name>
                          <Message.Username>
                    ....
```

vishakha-aggarwal:
Hey @sidmohanty11 I have made the required changes. Please review them. Thanks!

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "embeddedchat",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
sidmohanty11, line 3: can you undo this package-lock changes and I think this will be good to go.

sidmohanty11 (changes requested):

package-lock.json
vishakha-aggarwal: Done! Please have a look.

sidmohanty11 (approved):

PR #90, opened 2023-01-17
Support to see the members of the channel

### LINKED ISSUE #38: Support to see who all are present in the channel
As an EmbedChat developer

**I need to**: Add an option for the users to see all the members of that channel

**So That**: People know who all are there on the channel

**Acceptance Criteria**

- [ ] Create an option in more options called "members" to check out all members of the channel

imf-ali:
As an EmbedChat developer

##I need to:  Add an option for the users to see all the members of that channel

##So That: People know who all are there on the channel

- [x]  Create an option in more options called "members" to check out all members of the channel

Fixes #38 

## Video/Screenshots
[screen-capture.webm](https://user-images.githubusercontent.com/42451210/212967218-a0417a9f-ab91-4379-af6e-41bc04cba326.webm)

### DIFF
--- a/src/components/RoomMembers/RoomMember.js
+++ b/src/components/RoomMembers/RoomMember.js
@@ -8,9 +8,6 @@ import { useMemberStore } from '../../store';
 const RoomMembers = ({ members }) => {
   const toggleShowMembers = useMemberStore((state) => state.toggleShowMembers);
 
-  const showChannelMembers = async () => {
-    toggleShowMembers();
-  };
   return (
     <Box className={classes.modal} p="x16">
       <Box
@@ -27,7 +24,7 @@ const RoomMembers = ({ members }) => {
         <Icon name="members" pi="x2" size="x24" />
         <Box>Members</Box>
         <ActionButton
-          onClick={showChannelMembers}
+          onClick={toggleShowMembers}
           ghost
           display="inline"
           square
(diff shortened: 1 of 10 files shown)


src/store/memberStore.js
```diff
@@ -0,0 +1,10 @@
+import create from 'zustand';
+
+const useMemberStore = create((set) => ({
+  members: [],
+  showMembers: false,
+  showMembersHandler: (filter) => set(() => ({ showMembers: filter })),
```
sidmohanty11, line 6: As we are using this function for toggling the state only, can we do it over here? Something like `toggleShowMembers` so that you can remove functions where we have set it to be `false` or `true` and just use toggle? What do you think?

sidmohanty11 (changes requested):
Awesome work @imf-ali ! Please take a look at my comments

src/components/RoomMembers/RoomMember.js
```diff
@@ -0,0 +1,50 @@
...
+        is="h3"
+        pi="x24"
+        borderBlockEndWidth="default"
+        borderBlockColor="extra-light"
+        flexShrink={0}
+      >
+        <Icon name="members" pi="x2" size="x24" />
+        <Box>Members</Box>
+        <ActionButton
+          onClick={showChannelMembers}
```
sidmohanty11, line 30: I think we can directly pass `toggleShowMembers` here?

sidmohanty11 (changes requested):

sidmohanty11 (approved):
LGTM! Thanks a lot @imf-ali . Awesome work 🥳

PR #104, opened 2023-01-24
Add options to send messages in different formatted style

imf-ali:
As an EmbedChat developer

**I need to**: Different options available while sending a message. 
This will consist of following options:
1. Bold
2. Italic
3. Strike
4. Code
5. Multiline

**So That**: I can send in specified way as mention above

**Acceptance Criteria**

- [x] Able to make message bold
- [x] Able to italic text
- [x] Able to strike text
- [x] Able to include code
- [x] Able to new line
- [x] Setup markdown used in RC core repo
- [x] Make message box height dynamic to certain height

Fix #98 
Fix #10 

[screen-capture.webm](https://user-images.githubusercontent.com/42451210/214566407-c175e7be-3366-4322-b602-746b983bec6b.webm)

PR #109, opened 2023-01-28
Messages should be grouped together for same user and if sent consecutively (same/near time)

### LINKED ISSUE #106: Messages of the same user in same time frame should have single message header
As an EmbedChat developer

**I need to**: Have same message header for messages of same/near time by same user

**So That**: Chat body looks cleaner

**Acceptance Criteria**

- [ ] Multiple messages should have a single header if they fall within same/near time

**Expected behaviour**
image

**Actual behaviour**
image

imf-ali:
As an EmbedChat developer

**I need to**: Have same message header for messages of same/near time by same user

**So That**: Chat body looks cleaner

**Acceptance Criteria**

- [ ] Multiple messages should have a single header if they fall within same/near time

image

Fix #106


imf-ali:
Hey @sidmohanty11, there are few styling fix I need to do, but it would be better if I do those after this PR #104 is merged. Thank you.

sidmohanty11 (changes requested):
Hey @imf-ali thanks for the PR, it looks good to me. Could you please take a look at the merge conflicts

sidmohanty11 (approved):
Thanks @imf-ali ! LGTM

PR #130, opened 2023-02-13
Create .gitpod.yml

itsmanojthapa:
# this will start app automatically in gitpod


sidmohanty11:
@itsmanojthapa Thanks for this!! We really need to gitpod-ize EmbeddedChat soon! However, I had an approach in mind:

EmbeddedChat requires:
- An RC instance
- Playground app

So, we need a way to gitpod-ize it in such a way that,
- it spins up a dockerized RC app
- use that RC app's host URL to instantiate EC playground app

I'm open to suggestions, let me know what you think

itsmanojthapa:
@sidmohanty11 yes sure

sidmohanty11:
@itsmanojthapa yes two servers:
- one dockerized RocketChat server
- one EmbeddedChat dev server (`npm run dev`)

itsmanojthapa:
#182

PR #133, opened 2023-02-15
fix:changed hardcoded url to RCInstance.getHost() (RocketChat#132)

CulturalProfessor:
## Video/Screenshots
Screenshot from 2023-02-15 21-15-01

### DIFF
--- a/src/components/Attachments/Attachment.js
+++ b/src/components/Attachments/Attachment.js
@@ -1,12 +1,14 @@
 import { Box, Icon } from '@rocket.chat/fuselage';
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import ImageAttachment from './ImageAttachment';
 import AudioAttachment from './AudioAttachment';
 import VideoAttachment from './VideoAttachment';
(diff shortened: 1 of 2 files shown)


src/components/Attachments/Attachment.js
```diff
@@ -1,12 +1,14 @@
...
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import ImageAttachment from './ImageAttachment';
 import AudioAttachment from './AudioAttachment';
 import VideoAttachment from './VideoAttachment';
 import PinnedAttachment from './PinnedAttachment';
+import RCContext from '../../context/RCInstance';
 
 const Attachment = ({ attachment, host }) => {
```
sidmohanty11, line 10: Let's use `host` from here directly

sidmohanty11 (changes requested):
Thanks @CulturalProfessor for the PR!

src/components/Attachments/Attachment.js
CulturalProfessor: Ok I will use **host** directly from here on.

sidmohanty11 (approved):

Issue #132, opened 2023-02-14
UI and Code refactoring for release v0.0.3


sidmohanty11:
### UI
![image](https://user-images.githubusercontent.com/73601258/218664532-7d1887cc-a575-4d95-93b0-0803d4e6f361.png)

These are just my suggestions, if you have any amazing UI/UX idea please feel free to suggest it

### Code
- [x] Multiple API requests being made (refer #112) -> occurring at ChatBody.js while calling getMessagesAndRoles() twice
- [ ] Circular dependencies error from terminal `[0] (!) Circular dependencies
[0] src/components/Markup/elements/ItalicSpan.js -> src/components/Markup/elements/BoldSpan.js -> src/components/Markup/elements/ItalicSpan.js
[0] src/components/Markup/elements/BoldSpan.js -> src/components/Markup/elements/StrikeSpan.js -> src/components/Markup/elements/BoldSpan.js
[0] src/components/Markup/elements/ItalicSpan.js -> src/components/Markup/elements/BoldSpan.js -> src/components/Markup/elements/StrikeSpan.js -> src/components/Markup/elements/ItalicSpan.js`
- [ ] `src/components/auth` -> unnecessary Login.css, Login.module.css has improper casing - same with LoginForm, TwoFactorTotpModal.module.css
- [x] `src/components/Attachments/Attachment` has reference to hard-coded value to localhost - use RCInstance.getHost()
- [ ] `src/components/ChatHeader` - unused variables, do we need them?
- [x] `src/components/MessageList` -> `key` is missing
- [ ] `src/components/SearchMessage` unused variables
- [ ] `src/components/TotpModal` unused import
- [ ] Refactor three useEffects inside `src/index.js`
- [ ] Sometimes while development its giving a huge CSS loader error, does anybody know why?
- [ ] Ensure we use ESLint throughout the codebase and not disable rules (we can disable those which are not in our hands, for e.g, _id - dangling underscore) to get benefit of linting

VipinDevelops:
Hi, could you please assign me the issue of Multiple API requests being made and the circular dependencies error mentioned in the terminal? Additionally, I noticed some other potential issues in the codebase such as unnecessary files, hard-coded values, unused variables and imports, and missing keys. I would be happy to work on those as well. Lastly, can we make sure to use ESLint throughout the codebase and avoid disabling rules, except for those outside of our control like dangling underscores 
 Thanks!

VipinDevelops:
[...]

PR #140, opened 2023-02-18
 Fix issue of Multiple API Calls 

### LINKED ISSUE #112: APIs are calling more than once
### Description:

Whenever any user performs any functionality in EmbeddedChat, the getMessage API is called more than once which eventually results in status code 429 due to multiple requests. This issue might act as a blocker in implementing more features in the future.

### Steps to reproduce:

1. Send any message or react on any message
2. Do this multiple times

### Expected behavior:

The getMessage API should be called only once instead of calling itself multiple times.

### Actual behavior:

The API is calling multiple times due to which after some operations we get status code 429.

VipinDevelops:
# Fixing the issue of getMessage API being called multiple times

## Description
Whenever any user performs any functionality in EmbeddedChat, the getMessage API is called more than once which eventually results in status code 429 due to multiple requests. This issue might act as a blocker in implementing more features in the future. This PR aims to fix this issue by modifying the real-time function in api.js to only call the API once when a user sends a message or reacts to a message.

The getMessage API is only called once when a user sends a message or reacts to a message
Multiple requests are prevented, avoiding the status code 429 error
No regressions are introduced

## Changes Made
Modified the real-time function in api.js to use a Promise that resolves the first time the onMessage event is triggered, ensuring the getMessage API is only called once.

Fixes 
#112  and #132  Task 1 
- [x]  Multiple API requests being made 

## Video 
[Screencast from 2023-02-18 11-31-49.webm](https://user-images.githubusercontent.com/99081689/219844303-e359e1e6-1c85-4fc6-848e-0aa9fc9f58c3.webm)


VipinDevelops:
Hey, @sidmohanty11 I Fixed the issue of getMessage API being called multiple times resulting in status code 429.

I've thoroughly tested the solution and it's working as expected. Could you please take a look at the PR and let me know if you have any feedback or suggestions for improvement? I would greatly appreciate your review.

Thank you!!

src/lib/api.js
```diff
@@ -229,15 +229,25 @@ export default class RocketChatInstance {
...
       await this.rcClient.resume({ token: Cookies.get(RC_USER_TOKEN_COOKIE) });
-      await this.rcClient.subscribe('stream-room-messages', this.rid);
-      await this.rcClient.onMessage((data) => {
+      subscriptionId = await this.rcClient.subscribe(
+        'stream-room-messages',
+        this.rid
+      );
+
+      const messageHandler = (data) => {
+        this.rcClient.unsubscribe(subscriptionId);
```
sidmohanty11, line 242: If we are unsubscribing it, are we maintaining the realtime chat functionality?

src/lib/api.js
```diff
@@ -229,15 +229,25 @@ export default class RocketChatInstance {
...
         callback(data);
-      });
+      };
+
+      await this.rcClient.onMessage(messageHandler);
```
sidmohanty11, line 246: is `onMessage` room specific or called everytime?

sidmohanty11 (changes requested):
Thanks a lot @VipinDevelops for the PR! I had some questions

src/lib/api.js
VipinDevelops: Yes i showed that in preview also real-time communication is working perfectly fine

src/lib/api.js
VipinDevelops: it is called for all messages received by the client and  The streamDataHandler function is also not room-specific, but it filters events based on the room ID before calling the callback.

sidmohanty11:
@VipinDevelops this is not the right way in which we should handle it. I think @abhinavkrin and @vishakha-aggarwal mentioned a better way and we should proceed with that.

PR #142, opened 2023-02-18
Refactor RCComponent useEffects

VipinDevelops:

- [ ]  Combine three useEffect hooks into one
- [ ]  Set toast bar position and avatar state using store hooks
- [ ]  Check if the user is authenticated using the setIsUserAuthenticated hook
- [ ]    Verify that code behavior is not affected

Fixes #132  Task 9


abhinavkrin:
@VipinDevelops Thanks for this PR. I do have one question that should we fetch user details when props like toastPosition, showAvatar etc changes?
Tagging @sidmohanty11.

PR #156, opened 2023-02-28
[IMPROVE] shift+enter adds a newline in chat input

abhinavkrin:
# Shift+Enter adds a new line
Currently, only ctrl+enter adds a newline instead of rocket.chat in which both shift+enter and ctrl+enter add a newline. many users including me are used to the shift+enter way of adding new lines.

### DIFF
--- a/src/components/ChatInput/ChatInput.js
+++ b/src/components/ChatInput/ChatInput.js
@@ -115,6 +115,10 @@ const ChatInput = () => {
             }
           }}
           onKeyDown={(e) => {
+            if (e.shiftKey && e.keyCode === 13) {
+              // new line with shift enter. do nothing.
+              return;
+            }
             if (e.ctrlKey && e.keyCode === 13) {
               // Insert line break in text input field
               messageRef.current.value += '\n';


src/components/ChatInput/ChatInput.js
```diff
@@ -115,6 +115,10 @@ const ChatInput = () => {
             }
           }}
           onKeyDown={(e) => {
+            if (e.shiftKey && e.keyCode === 13) {
+              // new line with shift enter. do nothing.
```
sidmohanty11, line 119: do we need to add line breaks or not?

sidmohanty11 (changes requested):
Thanks @abhinavkrin me too!

src/components/ChatInput/ChatInput.js
abhinavkrin: The new line/carriage return character is already added to messageRef from the onChange event handler. The changes I made are in the onKeyDown handler.

sidmohanty11 (approved):

Issue #163, opened 2023-03-01
Section for tagging the user is not closing 


VenomFate-619:
### Description:
The section that is there for tagging the user is not closing despite removing the @ symbol also.
If I close and open the Embedded Chat, it is still there.

it mostly happens when the message box is empty,
and sometime when there is text in it.

### Steps to reproduce:

1. type @ in the empty message box
2. clear the @ 

### Actual behavior:

https://user-images.githubusercontent.com/67755128/222218450-c9ed5b0d-53aa-4850-8292-ecd0b5b87e83.mp4

VenomFate-619:
It consists of two task
- handling when textarea is empty
- handling when @ is typed after the text without the space (in this case there should be no suggestion) 
![image](https://user-images.githubusercontent.com/67755128/222223707-650b968d-5f2e-47bf-a68e-1c3ee5bca7d0.png)
no suggestion

![image](https://user-images.githubusercontent.com/67755128/222223831-667ea9d0-b3aa-4d0e-a708-2765838dbe3d.png)
suggestion

VenomFate-619:
@sidmohanty11 i like to fix this.

sidmohanty11:
Thanks @VenomFate-619 for picking this up!

raazseth:
Is this still open? If yes then please assign me this issue @sidmohanty11

PR #171, opened 2023-03-03
fix:tagging user section not closing

VenomFate-619:
# Brief Title
Whenever we type @ in empty textArea and erase it the list of the user to be mentioned still remains there.
The mention box should not be active when the user type @ after the text without the space

## Acceptance Criteria fulfilment

- [x] handling when textarea is empty
- [x] handling when @ is typed after the text without the space (in this case there should be no suggestion)

Fixes # (issue)
#163 
## Video/Screenshots

https://user-images.githubusercontent.com/67755128/222644141-828aa4bb-620e-4b1d-87d6-559195af1359.mp4

### DIFF
--- a/src/lib/searchToMentionUser.js
+++ b/src/lib/searchToMentionUser.js
@@ -7,39 +7,44 @@ export const searchToMentionUser = (
   setmentionIndex,
   setshowMembersList
 ) => {
-  if (message.length === 0) return;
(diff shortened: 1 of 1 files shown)


sidmohanty11 (approved):
Thanks @VenomFate-619 LGTM!

PR #174, opened 2023-03-04
[BUGFIX] mentions and roles causing initial app load to crash when unauthed

### LINKED ISSUE #173: Initital load crashes app when user is not authenticated.
### Description:

The app is crashing on the initial load when the user is unauthenticated.

### Steps to reproduce:

1. Make sure you are not already authenticated.
2. Run the app.

### Actual behavior:
error-mention

abhinavkrin:
# Brief Title
The app was crashing when the user is not authenticated. 
## Reasons:
1. trying to set mentions when the user mention data is not available. As the current implementations search if the user who is mentioned is present in the member list of the room or not. But we don't have that member list information when the user is authenticated.
2. Fetching roles (which requires auth) and calling functions on null/undefined objects.
- [x] No crash when user is not authenticated

Fixes #173

## Video/Screenshots
error-mention
error-getRoles

### DIFF
--- a/src/components/ChatBody/ChatBody.js
+++ b/src/components/ChatBody/ChatBody.js
@@ -29,16 +29,27 @@ const ChatBody = ({ height, anonymousMode, showRoles, GOOGLE_CLIENT_ID }) => {
   );
 
   const getMessagesAndRoles = useCallback(async (anonymousMode) => {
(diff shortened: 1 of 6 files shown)


sidmohanty11 (approved):
Thanks @abhinavkrin !

PR #190, opened 2023-03-12
[Fix: 132] Added markdown without circular dependency

vishakha-aggarwal:
### Acceptance Criteria fulfillment

- [x] Implemented markdown styles without circular dependency of components

Fixes #132 [Circular dependency]


vishakha-aggarwal:
Hey @sidmohanty11 Please review this PR. Thanks!

sidmohanty11 (changes requested):
@vishakha-aggarwal Thanks a lot for the PR! Looks like this approach needs too much prop drilling and chaining stuffs - which I don't think should be ideal and most probably redundant.

From what I can observe, we are requiring the classes only at the PlainSpan level and this makes code a bit unreadable to move with. Lets discuss a better approach for this and we  can update it here. What do you think?

PR #196, opened 2023-03-15
[IMPROVE] refactor-message-components

abhinavkrin:
# Brief Title
I have separated the message jsx from MesageList and ThreadMessage list to a separate reusable Message component. This would be beneficial reducing code size, increasing code readability and also when we would be introducing theming.


sidmohanty11 (approved):
Great refactor @abhinavkrin !

PR #199, opened 2023-03-15
[IMPROVE] typing status support for rc 6.0+

### LINKED ISSUE #198: User typing status not shown for some RC hosts
### Description:
User typing status is not being shown for some RC hosts. I was running 6+.

abhinavkrin:
# Brief Title
This PR will add support to show typing user status in EC connected to rocketchat >6.0.
Until now EC was only listening for  `roomId/typing` events to get typing users updates.
After updating my RC to the latest version, user typing status events were not being received by our real-time DDP connection. 
After spending hours looking into the codes and working on the current version, I found that typing status is now sent through `roomId/user-activity` event.
So to support all the RC versions, I subscribed for both `roomId/typing` and `roomId/user-activity` events

Fixes #198

### subscribing event
typingsend
## event messages 
typing


sidmohanty11 (approved):
Thanks @abhinavkrin ! I think we should discuss this more for handling this carefully for other versions. Having some test cases to run for different versions to check all functionalities will be ideal I think - lets take this to the official channel.

PR #218, opened 2023-07-01
React bootstrap

abhinavkrin:
This PR bootstraps react package. I have used our new API library and Auth library for this package.
The storybook is also setup for react package.
After this is merged we will gradually move to a new component system by replacing fuselage components with new components one by one or in batches.

## Video/Screenshots
Screenshot from 2023-07-01 13-29-52


sidmohanty11 (commented):
Closing this as we already have these changes in #222

PR #219, opened 2023-07-01
EmbeddedChat enhancements

sidmohanty11:


sidmohanty11:
@abhinavkrin lets get this in! We can publish a stable release and after that go with further enhancements, what do you think?

abhinavkrin:
Yes, We should make embeddedchat stable. I think we have enough features/enhancements for now. There are lots of things that need to be fixed. We will fix them and release them as the next major version.

PR #247, opened 2023-10-04
Fix/emoji

KushalSr:
# Brief Title

- [x] Task 1
- [x] Task 2
- [x] Task 3

Fixes # (issue)

### DIFF
--- a/packages/react/src/components/Markdown/Markdown.js
+++ b/packages/react/src/components/Markdown/Markdown.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import './Markdown.css';
-import { MessageEmoji } from '@rocket.chat/fuselage';
+//import { MessageEmoji } from '@rocket.chat/fuselage';
+import {MessageEmoji} from '../EmojiPicker/EmojiPicker'
+
 import PropTypes from 'prop-types';
 import emojione from 'emoji-toolkit';
 import { Markup } from '../Markup/index';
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Markdown/Markdown.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import './Markdown.css';
-import { MessageEmoji } from '@rocket.chat/fuselage';
+//import { MessageEmoji } from '@rocket.chat/fuselage';
+import {MessageEmoji} from '../EmojiPicker/EmojiPicker'
```
abhinavkrin, line 4: We do not export MessageEmoji from EmojiPicker. It will cause errors. 
MessageEmoji is used to wrap big emoji messages. That is when we send a single emoji as message, it appears to be slightly bigger in size.

abhinavkrin (commented):
MessageEmoji is not exported from EmojiPicker. MessageEmoji are used to render those single emoji when a single emoji is sent as message and it appears to be slighter bigger in size.

abhinavkrin:
This PR will break the build as MessageEmoji is not exported from EmojiPicker.

packages/react/src/components/Markdown/Markdown.js
KushalSr: Hi @abhinavkrin , thanks for your input. I'll fix that ASAP

PR #289, opened 2023-10-25
Fix/#265 added RC_HOST as env variable

### LINKED ISSUE #265: Set up RC host url as environment variable in react storybook
### Description
The RocketChat Host URL is currently hardcoded as `http://localhost:3000` in the React Storybook configuration. However, it is essential to provide flexibility and configurability by allowing users to set the host URL via environment variables. This change will make it easier for users to adapt the configuration to their specific needs.

### Proposed Solution
We should modify the React Storybook configuration to read the RocketChat Host URL from an environment variable. By default, if the environment variable is not set, it should fallback to `http://localhost:3000`.
So that if developers want to use HTTP://localhost:5000 as host url,  then the storybook could be started using:
```
RC_HOST=http://localhost:5000 yarn storybook
```
or by setting up a `.env` file (it should be added to gitignore).

### Steps to Implement
1. Update the React Storybook configuration to use an environment variable for the RocketChat Host URL.
2. Provide clear documentation on how users can set the environment variable.
3. Ensure that when the environment variable is not set, it defaults to `http://localhost:3000`.
4. Change all the hardcoded host URLs to use environment variable.
5. Update readme accordingly.

### Expected Behavior
Users should be able to easily configure the RocketChat Host URL for React Storybook by setting an environment variable. When no environment variable is set, it should default to `http://localhost:3000`.

tanu-chahal:
# Added RC_HOST variable to set up custom host for Rocket Chat Server

- [x] Added STORYBOOK_RC_HOST as an environment variable to set up a custom host.
- [x] Updated README.md accordingly.

Fixes #265

## Screenshot
running-with-custom-rc-host-2

In the screenshot using the host as 4000

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,12 @@ yarn storybook
 
 Till now, storybook should be up and running. You can play around with EmbeddedChat and it's component and see changes in storybook in real-time.
 
+By default, the Rocket Chat server host is http://localhost:3000, in case you have Rocket Chat server running on any other port/server you can try set up custome url as ```STORYBOOK_RC_HOST=<your-custom-url>``` variable by creating a ``.env`` file in ```packages/react``` directory or simply run this commmand:
+
+```bash
+STORYBOOK_RC_HOST=<your-custom-url> yarn storybook
+```
+
 ### Working with api and auth packages.
 
 #### Starting auth dev environment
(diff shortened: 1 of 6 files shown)


abhinavkrin (approved):
LGTM!

PR #302, opened 2023-11-11
Added markdown shortcuts for bold and italic font style (ctrl+b,ctrl+i)

Yashraj7890:
Markdown shortcuts for bold and italic

The user can now style the font to bold and italic by using ctrl+b and ctrl+i shortcuts respectively.

Behavior:

Once the user presses any of the above mentioned shortcut keys after selecting some text, the text will be wrapped around "* *" or "_ _" to ensure that the text gets styled accordingly when the message is actually sent.
If the shortcut keys are pressed while no text is selected, the symbols will still appear but with the cursor between them.

The video below depicts the actual behavior:

https://github.com/RocketChat/EmbeddedChat/assets/98230836/86b34b9f-323d-43fe-955a-1df347edf8d0

### DIFF
--- a/packages/react/src/components/ChatInput/ChatInput.js
+++ b/packages/react/src/components/ChatInput/ChatInput.js
@@ -342,6 +342,30 @@ const ChatInput = () => {
                 // new line with shift enter. do nothing.
                 return;
               }
+              if (e.ctrlKey && e.key === 'i') {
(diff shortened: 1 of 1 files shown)


abhinavkrin (approved):
Thanks @Yashraj7890. This looks good to me.

PR #345, opened 2023-12-07
Implemented tooltips using MUI

Yashraj7890:
Successfully implemented tooltips by using Material UI with support for touch screen devices.

The video below displays the behavior:

https://github.com/RocketChat/EmbeddedChat/assets/98230836/8f37c314-aee2-4495-a015-2729591e755d

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,5 +16,10 @@
     "esbuild": "^0.17.19",
     "lerna": "^6.6.2",
     "typescript": "^5.1.3"
+  },
+  "dependencies": {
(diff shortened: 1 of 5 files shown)


Yashraj7890:
😅Got it.
@abhinavkrin I'm thinking on implementing the Tooltip component just like the MUI library.
 
Let us suppose we want a tooltip to displayed whenever we hover or press and hold over a particular component.
so the code would go like this :

<tooltip>
<>the desired component</>
</tooltip>

should I go ahead with this?

PR #348, opened 2023-12-07
Implemented custom Tooltips 

Yashraj7890:
Successfully implemented custom tooltips with support for touch and hold action on touch screen devices.

Usage :

import {Tooltip} from '../Tooltip';

```
<Tooltip text="desired text" position="desired position">
<></>// tooltip will be activated for this element
</Tooltip>

```

https://github.com/RocketChat/EmbeddedChat/assets/98230836/fb2fa865-da46-4e25-8f2f-1b40139d2556


abhinavkrin (changes requested):
Great Work! I have also added some suggestions.

Yashraj7890:
Hey @abhinavkrin I have modified the code to make it more modular so that more positions can be added to the tooltip and the styles of the tooltip and the tooltipArrow can be modified accordingly however using emotion css in this case will reduce the modularity as I will be bound to use template literals here for each modification that must be made to align the tooltip and the tooltip arrow accordingly which will also reduce the readability. 

(the tooltip arrow must be rotated, and the position must be changed when the tooltip position changes and the code has been modified in such a way that this behavior can be injected easily but using emotion will result out in use of multiple template literals for conditional rendering)



https://github.com/RocketChat/EmbeddedChat/assets/98230836/17043fa0-318f-4330-98b3-29448169855a

abhinavkrin (approved):
Thanks @Yashraj7890 , The code looks cleaner now and the arrow is also great.

PR #349, opened 2023-12-08
FEATURE: Added EmojiPreview component for typing Emojis using : shortcut

### LINKED ISSUE #105: able to add emoji in text using : shortcut
As an EmbedChat developer

**I need to**: 
Add emoji  without emoji picker, instead use the`: `shortcut
this feature is present in most the chat applications
**So That**:
user experience becomes better.

Tasks

- on typing `: `suggestion box should come
- suggestion should be changed according to the text written after : 

Something like this
image

Kamlesh72:
# Description 
Added a Component called EmojiPreview that provides shortcuts to type Emojis.
Easily Access Emojis using ":"

Fixes #105 

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/90441742/82b8473c-3366-45ca-87e1-5ff0becc9867


Kamlesh72:
Things yet to be added:
Pressing Enter should select the current emoji. 
Also pressing ArrowUp shouldn't move the cursor to the start of text.

abhinavkrin:
Thank you @Kamlesh72 for the incredible work! Your contributions are paving the way for us to develop a custom emoji picker. This is particularly beneficial since our current emoji picker package is quite large, adding about 1MB to our bundle size. With the emoji set from your PR, we can design a more efficient and tailored emoji picker that better suits our needs. Your efforts are greatly appreciated!

PR #355, opened 2023-12-16
Injected tooltips for messageToolbox and provided more customizability to the tooltip component.

Yashraj7890:
Applied tooltips for messageToolbox and provided more customizable options to the tooltip component as micro-adjustments may be needed in some cases .

`<Tooltip position="" text="" X=""  Y="" >
</Tooltip> `

here X is equivalent to translateX() in css
and Y is equivalent to translateY() in css

https://github.com/RocketChat/EmbeddedChat/assets/98230836/ebc03549-2d32-4c43-9363-ddb4eaa3d043


Yashraj7890:
Hey @abhinavkrin, this wasn't an issue so I just created a PR as only some minor changes were made.

Issue #361, opened 2023-12-23
Resolve warnings appearing in console


umangutkarsh:
### Description:
Some warnings can be resolved in the console tab of chrome-developer tools (If Browser === Google Chrome).

### Steps to reproduce:

1. Do `yarn storybook`.
2. Right click on screen -> Click on Inspect
3. Open the console tab. See the warnings

### Expected behavior:
These warnings can be resolved.

### Actual behavior:
[warn.webm](https://github.com/RocketChat/EmbeddedChat/assets/95426993/7e1c0aa9-9b9f-46d4-89dd-f70d20bdea56)

### Warnings to resolve.
- [ ] Failed prop type: Invalid prop `setClosableState` of type `boolean` supplied to `EmbeddedChat`, expected `function`.
- [ ] Failed prop type: Invalid prop `setClosableState` of type `boolean` supplied to `ChatHeader`, expected `function`.
- [ ] Received `true` for a non-boolean attribute `small`.
- [ ] Each child in a list should have a unique "key" prop.
- [ ] React does not recognize the `isEditing` prop on a DOM element.
- [ ] React does not recognize the `isPending` prop on a DOM element.

devanshkansagra:
I have setup the development environment, but storybook is unable to connect to localhost:3000. it gives me this type of error:
![image](https://github.com/RocketChat/EmbeddedChat/assets/125076549/7172510e-14ce-47e8-b518-803d57759b68)

umangutkarsh:
[...]

PR #364, opened 2023-12-23
Warnings 5,6 Resolved

umangutkarsh:
# Brief Title
Resolved warnings 5 and 6

## Warnings resolved:

- [X] React does not recognize the isEditing prop on a DOM element.
- [X] React does not recognize the isPending prop on a DOM element.

Addresses #361


umangutkarsh:
In this pr, I've removed the `isEditing` and `isPending` props in the Box component at line 162 in `Message.js`, since the Box component doesn't have those properties where it is implemented. Everything was working fine. If I shouldn't removed those props please let know.
Thanks.

PR #367, opened 2023-12-25
Typing-text fixed

umangutkarsh:
# Brief Title
This PR fixes the issue mentioned in #365 .

## Acceptance Criteria fulfillment

- [X] Issue fixed
- [X] Checks passed

Fixes #365 

## Video/Screenshots
[type2.webm](https://github.com/RocketChat/EmbeddedChat/assets/95426993/4d95ae12-0b42-4fb2-bfeb-b2da3ad7784e)

PR #381, opened 2024-01-02
Typing-display-fix

### LINKED ISSUE #372: UX fix of display text when user is typing
### Description:
To fix the display of text, when a user is typing. It should be displayed only when others are typing. 
Not to display when user1 is typing. It should only be displayed to `user1` when `user2` is typing. 

Example-- Should be displayed `<user2> is typing...` to user1 and Vice-versa

### Steps to reproduce:

1. Go to Chat area.
2. Type some message
3. See your_name displayed as `<your_name> is typing...`

### Expected behavior:
Should only be displayed when others are typing.

### Actual behavior:

[type3.webm](https://github.com/RocketChat/EmbeddedChat/assets/95426993/bc77bac5-5085-439c-b5a1-0c8991ce7547)

umangutkarsh:
# Brief Title
This PR fixes the issue #372 , and the issue described in PR #367 (in comments)

- [X] Display of `<user> is typing...` text when other users are typing
- [X] Not to display to the current user when current user is typing.

Fixes #372 

## Video/Screenshots
[typing.webm](https://github.com/RocketChat/EmbeddedChat/assets/95426993/8b9370f5-7bb5-467d-b365-dc27fef07625)

### DIFF
--- a/packages/react/src/components/TypingUsers/index.js
+++ b/packages/react/src/components/TypingUsers/index.js
@@ -17,32 +17,84 @@ export default function TypingUsers() {
   }, [RCInstance, setTypingUsers, currentUserName]);
 
   const typingStatusMessage = useMemo(() => {
-    if (typingUsers.length === 0) return '';
-    if (typingUsers.length === 1)
+    if (!typingUsers.length) return '';
+
+    if (typingUsers.length) {
+      const otherTypingUsers = typingUsers.filter((user) => user !== currentUserName);
+
+      if (typingUsers.length === 1) {
+        if (typingUsers.includes(currentUserName)) {
+          return '';
+        }
+        return (
+          <span>
+            <b>{typingUsers[0]}</b>
+            {' is typing...'}
+          </span>
+        );
+      }
+
+      if (typingUsers.length === 2) {
+        if (typingUsers.includes(currentUserName)) {
+          return (
+            <span>
+              <b>{otherTypingUsers[0]}</b>
+              {' is typing...'}
+            </span>
+          );
+        }
+        return (
+          <span>
+            <b>{typingUsers[0]}</b>
+            {' and '}
+            <b>{typingUsers[1]}</b>
+            {' are typing...'}
+          </span>
+        );
+      }
+
+      if (typingUsers.length === 3) {
+        if (typingUsers.includes(currentUserName)) {
+          return (
+            <span>
+              <b>{otherTypingUsers[0]}</b>
+              {' and '}
+              <b>{otherTypingUsers[1]}</b>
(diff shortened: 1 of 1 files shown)


packages/react/src/components/TypingUsers/index.js
```diff
@@ -17,32 +17,84 @@ export default function TypingUsers() {
...
+              {' are typing...'}
+            </span>
+          );
+        }
+        return (
+          <span>
+            <b>{typingUsers[0]} </b>
+            {', '}
+            <b>{typingUsers[1]} </b>
+            {`and 1 more is typing...`}
```
abhinavkrin, line 72: The correct phrase is `A, B and 1 more are typing`. Therefore cases for otherTypingUsers.length === 3 and otherTypingUsers.length > 3 can have same logic.

packages/react/src/components/TypingUsers/index.js
```diff
@@ -17,32 +17,84 @@ export default function TypingUsers() {
...
-          {' is typing...'}
-        </span>
-      );
-    if (typingUsers.length === 2)
+    if (!typingUsers.length) return '';
+
+    if (typingUsers.length) {
+      const otherTypingUsers = typingUsers.filter((user) => user !== currentUserName);
+
+      if (typingUsers.length === 1) {
```
abhinavkrin, line 25: When you have already filtered typingUsers, and stored the list in otherTypingUsers, the later variable should be used afterwards. However, I wonder why we need to refilter, when it is already done in useEffect. The issue might be somewhere else if its happening.

abhinavkrin (changes requested):

packages/react/src/components/TypingUsers/index.js
umangutkarsh: Hi @abhinavkrin . Actually, the previous approach wasn't giving the correct output, which should be like - **not to show that the "current user" is typing, and show it only when others are typing**.

So, I had to resolve to use this manual approach of using if-else which covers all the cases. It works fine btw.

packages/react/src/components/TypingUsers/index.js
umangutkarsh: Ya, I'll change it, My bad.

umangutkarsh: But I think separating out the code for typingUsers === 3 and typingUsers > 3 would be more easy to read.

packages/react/src/components/TypingUsers/index.js
nilaygit-10721: can i work on this and resolve this?

PR #382, opened 2024-01-02
Add toggle password visibility

### LINKED ISSUE #369: [Accessibility] : Ability to toggle Password Visibility
As an EmbeddedChat developer

**I need to**:
I think we  should add the eye icon to toggle the encryption of password (which is very common in a lot of logging in interfaces.)
currently it looks like this 
image

**So That**:
Adding the eye icon, will be helpful if anyone mistypes the password or want to double check rather than typing the password again from zero
**Acceptance Criteria**

- [x] Add the eye icon (make sure its responsive)
- [x] Add the toggling logic to it

Akshun-01:
# Feature to toggle between Password Visibility
I added the eye icon (closed and open) to toggle between password visibility and added logic to ensure proper functioning!

- [x] Add the eye icon (make sure its responsive)
- [x] Add the toggling logic to it

Fixes #369 

## Video/Screenshots
image

image

### DIFF
--- a/packages/react/src/components/auth/LoginForm.js
+++ b/packages/react/src/components/auth/LoginForm.js
@@ -10,6 +10,7 @@ import { Input } from '../Input';
 export default function LoginForm() {
   const [userOrEmail, setuserOrEmail] = useState(null);
   const [password, setpassword] = useState(null);
+  const [showPassword, setShowPassword] = useState(false);
   const isLoginModalOpen = loginModalStore((state) => state.isLoginModalOpen);
   const setIsLoginModalOpen = loginModalStore(
     (state) => state.setIsLoginModalOpen
@@ -28,6 +29,9 @@ export default function LoginForm() {
   };
   const handleEditPassword = (e) => {
     setpassword(e.target.value);
+  };
+  const handleTogglePassword = () => {
+    setShowPassword(!showPassword);
   };
 
   const fieldCSS = css`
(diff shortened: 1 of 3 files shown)


packages/react/src/components/auth/LoginForm.js
```diff
@@ -87,7 +97,49 @@ export default function LoginForm() {
...
+                type={showPassword ? 'text' : 'password'}
+                onChange={handleEditPassword}
+              />
+              <Box
+                type="button"
+                css={passwordEyeCss}
+                onClick={handleTogglePassword}
+              >
+                {showPassword ? (
+                  <svg
```
sidmohanty11, line 110: Lets move the svgs to a different file

sidmohanty11 (commented):
Thanks @Akshun-01 ! Looks awesome, left a minor comment

packages/react/src/components/auth/LoginForm.js
Akshun-01: hi, do we have any common assets folder or some other rules for adding svg's ?

Akshun-01:
Also @sidmohanty11 , could please check the comment on the main issue, I added some other things that can be added to login. Should I open different issues for each, or continue adding here only (new feature with each commit) ?

packages/react/src/components/auth/LoginForm.js
Akshun-01: @sidmohanty11, can you please clarify so that I can move forward with this issue.
Thanks!

Akshun-01:
Hello @sidmohanty11 , looked around and found how other svg's are exported. Did the same with these.
Could you please review the changes! ^^

sidmohanty11 (approved):
Thanks @Akshun-01 !

PR #409, opened 2024-01-18
Injected tooltips for message toolbox and added more customizability to the tooltip component

Yashraj7890:
Applied tooltips to messageToolbox and also added more customizability options to the component which may be needed 
when the component overflows out of the viewport or to adjust it's relative position, also slightly improved it's Ui.

<Tooltip X=""  Y="" >

</Tooltip>

where X is equivalent to translateX and Y is equivalent to translateY in css.

Screenshot (54)


Yashraj7890:
Some merge conflicts that came up after the PR was created have been resolved.

Yashraj7890:
Hey @sidmohanty11 some changes were made to the tooltip component recently and those changes tried fixing the tooltip component's positioning which worked partially. I have modified the tooltip component and now it works perfectly also some merge conflicts that came up after the PR was created have been resolved.

PR #426, opened 2024-01-24
Fixed room information component

### LINKED ISSUE #425: Unnecessary appearance of scrollbar in room information component and component not hiding away when clicked outside the component body.
### Description:

The room information component displays scrollbars that are not needed, and the component is not disappearing when clicked outside the body of the component.

Screenshot (58)

Yashraj7890:
# Brief Title

 Fixed the component's functionality to hide when clicked outside the component and also removed the unwanted scrollbars.

Fixes # (425)

Closes #425

### DIFF
--- a/packages/react/src/components/RoomInformation/RoomInformation.js
+++ b/packages/react/src/components/RoomInformation/RoomInformation.js
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
 import { css } from '@emotion/react';
 import { Avatar } from '../Avatar/Avatar';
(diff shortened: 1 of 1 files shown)


Spiral-Memory:
Good work, @Yashraj7890 . I would like to add a few points here. Ensure that hiding the scrollbar does not affect the scrollbar in the Threads menu.

Also, closing the menu after clicking outside is not an appropriate behavior in this case. If it were designed with an overlay over the chat body, that would be expected behavior. However, currently, both the chat body and menu is expected and designed to work at the same time.

So hiding the scrollbar is good, but the rest I feel is not appropriate in my opinion.

nilaygit-10721:
can i work on this ?

PR #442, opened 2024-02-01
feat: Added a Threads menu to easily access ongoing threads in a channel.

### LINKED ISSUE #441: [Feat]: Adding a section to view all the threads going in the channel
As an EmbeddedChat developer, 

**I need to**:
Add a section where all threads can be seen in one place.

**So That**:
Users can easily navigate through the ongoing threads in the channel.

**Acceptance Criteria:**

- [ ] Identify the modal where these threads should be displayed.
- [ ] Design the modal in a manner consistent with the existing layout of the Rocket.Chat main application.
- [ ] Retrieve thread information using the REST API and present it in the modal. Each thread should be clickable, allowing users to navigate directly to the respective thread.

For reference, please find the video of the original Rocket.Chat application, which inspired this idea:

https://github.com/RocketChat/EmbeddedChat/assets/78961432/03e9f58a-427b-4008-b290-0335bd6277f1

Spiral-Memory:
# Brief Title
Added a new section "Threads menu" where users can navigate through ongoing threads in the channel, search for different messages, and navigate directly to those threads.

- [x] Identified the correct modal in which we can display ongoing threads.
- [x] Designed it consistently with the original Rocket.Chat design and reused existing components in the embedded chat.
- [x] Since the message was already retrieved, used the global message state and filtered through the messages that are threads and not just messages or other notifications.
- [x] Each thread in the menu is clickable, allowing users to navigate directly to that thread.
- [x] Threads are searchable, enabling users to easily search through thread messages.
- [x] Created a UI for "No Thread Messages" to inform users when there are no threads that match the search.
- [x] Also checked the `showAvatar` state that can be given as props and will be disabled when set to false.

Fixes #441 

## Video/Screenshots
The following video demonstrates how it works:

https://github.com/RocketChat/EmbeddedChat/assets/78961432/9ef0a24f-46e0-4089-9cb5-457e5ff89f05

The following screenshot shows that the avatar will be displayed only when enabled:
Screenshot from 2024-02-01 15-42-19

### DIFF
--- a/packages/react/src/components/AllThreads/AllThreads.js
+++ b/packages/react/src/components/AllThreads/AllThreads.js
@@ -1,0 +1,149 @@
+import React, { useState, useMemo } from 'react';
+import { css } from '@emotion/react';
+import classes from './AllThreads.module.css';
+import { Icon } from '../Icon';
+import { Box } from '../Box';
+import { ActionButton } from '../ActionButton';
+import { useMessageStore, useUserStore, useThreadsMessageStore } from '../../store';
+import { MessageBody } from '../Message/MessageBody';
+import { MessageMetrics } from '../Message/MessageMetrics';
+import MessageAvatarContainer from '../Message/MessageAvatarContainer';
+import MessageBodyContainer from '../Message/MessageBodyContainer';
+import MessageHeader from '../Message/MessageHeader';
+
+
+const MessageCss = css`
+  display: flex;
+  flex-direction: row;
+  align-items: flex-start;
+  padding-top: 0.5rem;
+  -webkit-padding-before: 0.5rem;
+  padding-block-start: 0.5rem;
+  padding-bottom: 0.25rem;
+  -webkit-padding-after: 0.25rem;
+  padding-block-end: 0.25rem;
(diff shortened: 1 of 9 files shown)


Akshun-01:
Great work! 🙌
- I want to point out that it should be confined to the chat area only.?
(Like starting from the header till where the message box ends.)
Because this may it might cause issue when integrated into other apps, right?

Spiral-Memory:
> Great work! 🙌
> - I want to point out that it should be confined to the chat area only.?
> (Like starting from the header till where the message box ends.)
> Because this may it might cause issue when integrated into other apps, right?

Actually, I had a similar thought while working on this, but currently, modals are designed that way. In another issue, my plan is that once the sidebar opens, the chat body and chat header will shrink to the other side, preventing the sidebar from overlapping. I'm already working on this and will soon discuss it with the maintainers and will raise another PR. Thank you for your feedback and appreciation, @Akshun-01

abhinavkrin (approved):
Hey @Spiral-Memory That's an impressive work. Thanks for adding this, it was in our todo list since a long time.

PR #462, opened 2024-02-12
fix: fixed logout issue: calling logout api successfully

### LINKED ISSUE #461: [BUG]: Unable to log out a user in reality; it only deletes the ec_token but never invalidates the authToken on the RC server.
### Description:

In Embedded Chat, the system is unable to log out the user, and the API call to /api/v1/logout fails. This issue results in the continued receipt of messages even after the user has logged out.

### Steps to reproduce:

1. Set up the embedded chat and log in.
2. Log out from the embedded chat and remain on the login screen.
3. Send a message from the Rocket Chat main app, and the embedded chat will still receive messages.
4. Additionally, one can view the console log to observe the issue: "Fails to logout."

### Expected behavior:

1. The authToken must be invalidated on the RC server as well by calling the logout endpoint.

### Actual behavior:

1. AuthToken not invalidated on the server.
2. Able to receive messages even after logout.
3. Console errors show that it fails to logout.

https://github.com/RocketChat/EmbeddedChat/assets/78961432/7ccd737f-9a60-452b-90d4-8a35028bbfee

Spiral-Memory:
# Brief Title
The logout issue has been resolved, and users can now log out successfully after calling logout api. Messages will no longer be received after logging out.

- [x] Debugged the issue and identified that the problem lies in sending headers to the REST API. The AuthToken and userId were not being sent in the header, even though the logout API requires authentication.
- [x] Discovered that the `getFetchConfig` function adds headers to the configuration, including 'Content-Type': 'application/json'. Now, we can directly send `...config.headers`, which will always contain the content-type header and any additional headers required.

Fixes #461 

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/78961432/24ac0c27-a554-48e4-847d-81f08f984928

Note: Rebuilding is required to reflect these changes in the application.

## Additional Information 

Although the logout issue has been fixed, another issue has been identified in the Embedded Chat if we correctly implement logout functionality maybe related with Rocket Chat SDK, specifically in the `resume()` function. After logging out, sending an updated token to the RC server using the `resume` function results in a successful connection. However, along with it also returns an unhandled promise error if we login without refreshing, however it doesn't affect the application.

I have tested this behavior and suspect the issue occurs after running the `resume` function in the `connect()` function in `EmbeddedChatApi`. Despite a try-catch block, the returned promise remains unhandled for reasons I am currently unaware of. This issue only arises when a user logs out and attempts to log in again without refreshing the page. Screenshots and a video explaining the problem are attached.

Code snippet where the issue might be occurring in the `resume` function:

Code Image

Video of the issue:

https://github.com/RocketChat/EmbeddedChat/assets/78961432/fa28d254-52db-415b-8c4e-ec6c73ebc6d5

Image of the issue:

Issue Image

Similar Issue that has been already raised:

- [Rocket.Chat.iOS Issue](https://github.com/RocketChat/Rocket.Chat.iOS/issues/1983)
- [Forum Discussion](https://forums.rocket.chat/t/user-logged-out-of-server-rest-api-iframe-commands/1631)

### DIFF
--- a/packages/auth/src/Api.ts
+++ b/packages/auth/src/Api.ts
@@ -33,7 +33,7 @@ export class Api {
 			body: data ? JSON.stringify(data) : undefined,
 			method,
 			headers: {
-				'Content-Type': 'application/json'
+				...config.headers,
 			}
 		});
 		if (!response.ok) {


Spiral-Memory:
I have fixed the logout issue as mentioned, but this helped me identify another issue with respect to similar thing which I have also explained in my PR. @abhinavkrin  @sidmohanty11 , if any of you are aware about why this issue arises, I would like to have a discussion regarding why this issue might arise.

abhinavkrin (approved):

PR #469, opened 2024-02-17
Setup basic e2e tests

sidmohanty11:
This PR adds an e2e suite using playwright that will help us reduce shipping broken builds and will make sure everything runs.


sidmohanty11:
@abhinavkrin merging this

PR #472, opened 2024-02-17
fix: linting issues

Spiral-Memory:
Fixing linting issues

### DIFF
--- a/packages/react/src/components/uiKit/blocks/UiKitModal.js
+++ b/packages/react/src/components/uiKit/blocks/UiKitModal.js
@@ -1,3 +1,6 @@
+/* eslint-disable react/jsx-no-undef */
+/* eslint-disable import/extensions */
+/* eslint-disable import/no-unresolved */
 /* eslint-disable no-shadow */
 /* eslint-disable react/jsx-no-constructed-context-values */
 import React from 'react';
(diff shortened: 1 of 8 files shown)


sidmohanty11 (approved):
Thanks @Spiral-Memory !

packages/react/src/components/uiKit/blocks/UiKitModal.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable react/jsx-no-undef */
```
abhinavkrin, line 1: Instead of disabling the errors. It will be better to comment out where the imported functions or components are being used and add label "to do" as comment.

abhinavkrin (changes requested):

packages/react/src/components/uiKit/blocks/UiKitModal.js
Spiral-Memory: Oops ! It's been merged now, what to do ? @abhinavkrin

abhinavkrin: No worries Raise another PR for this.

Spiral-Memory: Okay Sure

PR #479, opened 2024-02-22
add confirm modal before pinning messages

### LINKED ISSUE #450: Confirmation Before Pinning message
As an EmbeddedChat developer

**I need to**: 
Add a confirmation popup before pinning a message. 
image

**So That**:
The user knows pinned messages are visible for everyone to see. It increases the user experience.
And as it is an extension of RocketChat, it would make it more similar and consistent.

**Acceptance Criteria**
- [ ] Check pin status for message and give popup accordingly
- [ ] on confirmation pin the message otherwise return

Akshun-01:
# Implement Confirmation before Pinning Message feature

- [x] Check for the state to trigger the modal
- [x]  Make the confirmation Modal
- [x] Handle confirmation or canceling of the modal

Fixes #450 

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/93793691/fb971e87-7d1d-494b-8140-5d0d46fa8259

### DIFF
--- a/packages/react/src/components/Message/Message.js
+++ b/packages/react/src/components/Message/Message.js
@@ -1,4 +1,4 @@
-import React, { memo, useContext, useMemo } from 'react';
+import React, { memo, useContext, useMemo, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { format } from 'date-fns';
 import { css } from '@emotion/react';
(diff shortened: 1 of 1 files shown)


sidmohanty11 (commented):
Thanks for the PR @Akshun-01 but I'm not sure if this addition is necessary

Akshun-01:
Yeah, Zishan and I also discussed the same. 
I could only think of one use case where this might be important or necessary. 
Integration in a professional platform where you can't go on pinning stuff by mistake as it is public and can't be deleted.
But yeah, a very niche use case and can be ignored in most cases. 
So, we can leave the feature out!

PR #500, opened 2024-03-07
Added no pinned messages and no starred messages text in the message list component.

### LINKED ISSUE #499: Blank appearance of the message list after filtering pinned or starred messages (when no message is pinned or starred)
### Description:

If we go the pinned/starred messages section when no message has been pinned or starred then we have blank chat body, instead it would be better to have a simple message like in rocket.chat which would be more appropriate for the user.

Screenshot (79)

### Actual behavior:
Screenshot (73)

Yashraj7890:
After Changes:

Screenshot (75)

Screenshot (76)

Fixes #499

### DIFF
--- a/packages/react/src/components/MessageList/MessageList.js
+++ b/packages/react/src/components/MessageList/MessageList.js
@@ -7,8 +7,6 @@ import {
   useSearchMessageStore,
   useChannelStore,
   useUserStore,
-  useMentionsStore,
-  useThreadsMessageStore,
 } from '../../store';
 import RoomMembers from '../RoomMembers/RoomMember';
 import MessageReportWindow from '../ReportMessage/MessageReportWindow';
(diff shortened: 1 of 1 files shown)


sidmohanty11 (approved):

packages/react/src/components/MessageList/MessageList.js
```diff
@@ -7,17 +7,17 @@ import {
...
 import MessageReportWindow from '../ReportMessage/MessageReportWindow';
 import isMessageSequential from '../../lib/isMessageSequential';
 import SearchMessage from '../SearchMessage/SearchMessage';
 import Roominfo from '../RoomInformation/RoomInformation';
 import AllThreads from '../AllThreads/AllThreads';
-import UserMentions from '../UserMentions/UserMentions';
 import { Message } from '../Message';
+import { Icon } from '../Icon';
+
+import useThreadsMessageStore from '../../store/threadsMessageStore';
```
sidmohanty11, line 20: why remove the import like: {useThreadsMessageStore} from '../../store' ?

packages/react/src/components/MessageList/MessageList.js
```diff
@@ -61,7 +76,6 @@ const MessageList = ({ messages }) => {
       {showSearch && <SearchMessage />}
       {showChannelinfo && <Roominfo />}
       {showAllThreads && <AllThreads />}
-      {showMentions && <UserMentions />}
```
sidmohanty11, line 64: is there a need to remove usermentions?

Yashraj7890:
Apologies for the carelessness @sidmohanty11, that may have happened because of some experimentation which I was doing while writing the code. It has been taken care of and I have also added the same functionality for the room messages too :

Screenshot (80)

sidmohanty11 (approved):
Thanks @Yashraj7890 !

PR #510, opened 2024-03-11
chore: fixed message send issue using send-btn

Spiral-Memory:
# Brief Title
Fixed message sending issue using the send-btn. The send-btn was opening long Message modal instead of sending message

- [x] Properly calling the sendMessage function
- [x] Fixing namings

Fixes NA

### DIFF
--- a/packages/react/src/components/ChatInput/ChatInput.js
+++ b/packages/react/src/components/ChatInput/ChatInput.js
@@ -95,7 +95,7 @@ const ChatInput = ({ scrollToBottom }) => {
     (state) => state.setIsLoginModalOpen
   );
 
-  const [msgLongMsgLongErrorModal, setMsgLongErrorModal] = useState(false);
+  const [isMsgLong, setIsMsgLong] = useState(false);
 
   const {
     editMessage,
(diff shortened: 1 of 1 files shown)


packages/react/src/components/ChatInput/ChatInput.js
```diff
@@ -548,7 +548,7 @@ const ChatInput = ({ scrollToBottom }) => {
               <ActionButton
                 ghost
                 size="medium"
-                onClick={sendMessage}
```
Spiral-Memory, line 551: This is the cause of the issue

packages/react/src/components/ChatInput/ChatInput.js
```diff
@@ -548,7 +548,7 @@ const ChatInput = ({ scrollToBottom }) => {
...
+                onClick={() => sendMessage()}
```
Spiral-Memory, line 551: Resolution here ...

Spiral-Memory:
I introduced a small bug in my previous PR, fixing it here...

sidmohanty11 (approved):

PR #512, opened 2024-03-12
[Feat] Add a "Scroll to Top" button

thesynthax:
# Brief Title
Feature: Adding a "Scroll to Top" button to go to the start of conversation

- [x] Added the button as shown in the video below

Fixes Issue RocketChat/Rocket.Chat#511

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/29705333/8758b14e-3f14-4b5c-a4ea-1b38a7e17463

### DIFF
--- a/packages/react/src/components/ChatHeader/ChatHeader.js
+++ b/packages/react/src/components/ChatHeader/ChatHeader.js
@@ -34,6 +34,7 @@ const ChatHeader = ({
   styles = {},
   anonymousMode,
   showRoles,
+  messageListRef,
 }) => {
   const { classNames, styleOverrides } = useComponentOverrides('ChatHeader');
   const channelInfo = useChannelStore((state) => state.channelInfo);
@@ -97,6 +98,14 @@ const ChatHeader = ({
       setIsUserAuthenticated(false);
     }
   }, [RCInstance, setIsUserAuthenticated]);
+
+  const scrollToTop = useCallback(() => {
+    if (messageListRef && messageListRef.current) {
+      requestAnimationFrame(() => {
+        messageListRef.current.scrollTop = -messageListRef.current.scrollHeight;
+      });
+    }
+  }, [messageListRef]);
 
   const showStarredMessage = useCallback(async () => {
     const { messages } = await RCInstance.getStarredMessages();
@@ -266,6 +275,12 @@ const ChatHeader = ({
             label: 'Room Information',
             icon: 'info',
           },
+          {
+            id: 'scroll-to-top',
+            action: scrollToTop,
+            label: 'Scroll to Top',
+            icon: 'arrow-up',
+          },
         ]
       );
     }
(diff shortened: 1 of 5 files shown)


Spiral-Memory:
Good work! However, I don't see the necessity for this feature. In any chatting app, I have never encountered such a feature because there isn't much use case for it. Imagine a scenario with thousands of messages; does it make sense to go to the top message? Additionally, the placement of this feature is incorrect. It should not be in the kebab menu, options in this menu are used for opening sidebars here. Perhaps adding the option to go to the recent message could be considered, triggered by a floating down icon in chat body when users scroll up. However, this feature is not available in the main RC app. So, I don't think these features would be very useful.

thesynthax:
[This issue](https://github.com/RocketChat/Rocket.Chat/issues/31714) describes the necessity. I have personally faced situations in WhatsApp and Discord where it was required to reach the top. It allows to establish the context of the origin of the conversation. Even with the GSoC team channels, most of them had certain messages at the start that were crucial for the aspiring GSoC contributors. Other than that, I feel that this feature may come to use for many, if not all people. Talking about the positioning of the button, I will place it wherever as required or as suggested

Spiral-Memory:
I understand, but there is a reason why WhatsApp/Discord does not offer this feature. If there are numerous messages, when scrolling up, the application has to load the older messages each time. This process of reloading continues as you scroll up, making the overall experience very inefficient. Additionally, there is no control over how far back you want to go. A more effective feature would be to provide a date option, allowing users to select the specific date they want to navigate to. This gives users more control and streamlines the process. So, having the option to scroll down to go to recent message is one thing that can be added or providing a date picker for accessing older messages and selecting the desired time frame can be added.  Rest is up to you..

thesynthax:
That is indeed a good idea to go to a message by date, even Whatsapp has recently implemented this feature. And whatsapp mods already have this feature since long. Maybe integrating this scroll to top option into the date picker would be the best case for this feature.

Spiral-Memory:
Yep! But that also has to be efficient. It depends on how messages are categorized on the backend. If they are categorized by date, users can access key-value pairs and navigate to those messages efficiently. Many factors need to be taken care of, and it also depends heavily on how the Rocket Chat server handles it. So, I suggest first proposing this feature there. Once it is optimized for message navigation by date, it can be implemented here.

thesynthax:
Yes I understand

thesynthax:
Hey @sidmohanty11, @abhinavkrin. Please take a look at this PR whenever you are available. Thank you

KhalanderCSE734:
Hello @thesynthax, if the issue is not yet resolved, Can you please assign me this issue, I want to work on this issue.

Spiral-Memory:
Hey @KhalanderCSE734 ,

Thank you for your interest! This is not something we want to implement right now. You can check out other open issues. Thanks!

KhalanderCSE734:
Okay Thank you very much @Spiral-Memory

PR #514, opened 2024-03-12
MessageBox Responsive Fix

### LINKED ISSUE #513: MessageBox is Not Responsive for Minimize mode
### The Message Box is not Responsive for Minimized mode of EC

When we use `@` or `/` then the mention and command Modals take space and the message box disappears. This should not happen at least on the default screen sizes of EC (Minimized and Maximized mode)

### Steps to reproduce:

1. Use `@` or `/` to bring up mention or command menu
2. Observe the Message Box 

### Expected behavior:
Even on smaller screen Message Box should be visible.
image

### Actual behavior:
The Message Box disappears when modal is triggered.
image

Akshun-01:
# Make MessageBox Responsive and add arrow key iteration Feature to command Menu

- [x] Make the message box responsive so that the mention and command menu can be opened without disturbing it.
- [x] Make the mention menu scrollable such that the current item is shown in the modal window
- [x] Add Arrow iteration Feature to the command menu, similar to the command menu.

Fixes #513 

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/93793691/e626e799-66c7-48f3-8659-f9d2a5c8ca5d

### DIFF
--- a/packages/react/src/components/ChatBody/ChatBody.js
+++ b/packages/react/src/components/ChatBody/ChatBody.js
@@ -3,7 +3,18 @@ import React, { useCallback, useContext, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { css } from '@emotion/react';
 import RCContext from '../../context/RCInstance';
-import { useMessageStore, useUserStore, useChannelStore } from '../../store';
+import {
+  useMessageStore,
+  useUserStore,
+  useChannelStore,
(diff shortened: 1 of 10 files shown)


Akshun-01:
This currently has a small bug, on pressing `Enter` while in the command menu, it sends `/` as a message.
I am looking into it. Will fix this soon! ^^

PR #520, opened 2024-03-13
[Fix] Pin icon should not be visible to non-admin users

### LINKED ISSUE #519: [Feature Request] Pin icon should not be visible to users without pin permissions
Pin icon is not visible to users without pin permissions in the main Rocket.Chat application. This should be implemented in EmbeddedChat as well.  

**Acceptance Criteria**

- [ ] Pin icon to be not visible to users not having pin permissions.

ss_144

ss_145

thesynthax:
# Brief Title
Pin icon should not be visible to non-admin users like in main Rocket.Chat application.
Check #519 and #517 for complete details

- [ ] Pin Icon is now not visible to non-admin users

Fixes #519

## Video/Screenshots

ss_146

P.S: This PR is a continuation to from #518 as it requires the correct usage for UserStore.

### DIFF
--- a/packages/react-native/src/views/ChatRoomView/ChatRoomView.js
+++ b/packages/react-native/src/views/ChatRoomView/ChatRoomView.js
@@ -4,7 +4,7 @@ import { ChatInput } from '../../components/ChatInput';
 import { MessageActionsSheet } from '../../components/MessageActionsSheet';
 import { MessageList } from '../../components/MessageList';
 import { useRCContext } from '../../contexts/RCInstance';
-import { useMessageStore, useUserStore } from '../../store';
+import { useMessageStore, useChannelStore, useUserStore } from '../../store';
 
 const styles = StyleSheet.create({
 	container: {
@@ -16,7 +16,7 @@ const ChatRoomView = () => {
 	const { RCInstance, ECOptions } = useRCContext();
 
 	const isUserAuthenticated = useUserStore((state) => state.isUserAuthenticated);
-	const setRoles = useUserStore((state) => state.setRoles);
+	const setMemberRoles = useChannelStore((state) => state.setMemberRoles);
 
 	const setMessages = useMessageStore((state) => state.setMessages);
 	const upsertMessage = useMessageStore((state) => state.upsertMessage);
@@ -56,7 +56,7 @@ const ChatRoomView = () => {
 							Object.assign(obj, { [item.u.username]: item }),
 						{}
 					);
-					setRoles(rolesObj);
+					setMemberRoles(rolesObj);
 				}
 			} catch (e) {
 				console.error(e);
--- a/packages/react/src/components/EmbeddedChat.js
+++ b/packages/react/src/components/EmbeddedChat.js
@@ -116,6 +116,7 @@ const EmbeddedChat = ({
   );
   const setAuthenticatedUserId = useUserStore((state) => state.setUserId);
   const setAuthenticatedName = useUserStore((state) => state.setName);
+  const setAuthenticatedUserRoles = useUserStore((state) => state.setRoles);
 
   useEffect(() => {
     RCInstance.auth.onAuthChange((user) => {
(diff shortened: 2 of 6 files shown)


Spiral-Memory:
This is not a valid solution because it's not only the admin who can pin messages. The permissions to pin messages will have to be decided by the workspace administrator in Rocket.chat permission settings. Based on that, whoever has the permission to pin messages will be allowed. So first, you have to fetch the permission info, then check if the current person has permission to do so, and only after that can you make this decision. I am attaching a screenshot reference for you:

image

thesynthax:
Thanks a lot for this insight @Spiral-Memory, I will try to create a commit which will fetch the necessary permissions and check through those. Thanks again!

packages/react/src/components/ChatHeader/ChatHeader.js
```diff
@@ -181,7 +194,7 @@ const ChatHeader = ({
       if (res.success) {
         setChannelInfo(res.room);
         if (res.room.t === 'p') setIsChannelPrivate(true);
-        if (res.room.ro) setMessageAllowed();
+        if (!res.room.ro) setMsgAndPinAllowed();
```
Spiral-Memory: Why have you negated the res.room.ro ? It was made for read only room. It might break the existing functionality of when to set the messages. Please make a different function to hide the pin message icon.

thesynthax: I already fixed this in the last commit (fix readonly room [3935be0](https://github.com/RocketChat/EmbeddedChat/pull/520/commits/3935be06c498f5b14371e0136ff1c0bdd6b78764)), please check.
Also I already tried making a different function, but the API calls to get permissions were getting saturated very quickly as the API was called multiple times if a new function for pinning was made. So I decided that I can set canPinMsg from the same function itself.

Spiral-Memory: But now, how will disabling pinning messages work if the room is not set to read-only? Can you explain the flow ?

thesynthax: > Because this function `if (res.room.ro) setMsgAndPinAllowed();` will only be called when the room is read-only. What about normal rooms? Earlier, while setting whether sending msg is allowed, it depended on whether the room is read-only or not, but in your case, it does not.

Before my commit, the code was `if (res.room.ro) setMessageAllowed();`. That means this also ran only when the room was set to read-only. For normal rooms, canSendMsg variable was already initialized as true, so didn't need to run the function. Now I have just added the canPinMsg variable in the same function and renamed the function to setMsgAndPinAllowed.



> If there is an issue making two functions, what you can do is this: always call `setMsgAndPinAllowed` without any condition and pass `res.room.ro` as an argument. Then, in the `setMsgAndPinAllowed` function, only run the `setMsg` part if the room is read-only. Also, make sure you refactor the code accordingly.

You are right, will do it right away!

Spiral-Memory: Reply for Edit: Instead of closing this PR, you can close the other one and mark this as ready for review once everything looks good. In the PR description, mention that this fixes both issues so that both will get merged right away.

packages/react-native/src/views/ChatRoomView/ChatRoomView.js
```diff
@@ -16,7 +16,7 @@ const ChatRoomView = () => {
 	const { RCInstance, ECOptions } = useRCContext();
 
 	const isUserAuthenticated = useUserStore((state) => state.isUserAuthenticated);
-	const setRoles = useUserStore((state) => state.setRoles);
+	const setMemberRoles = useChannelStore((state) => state.setMemberRoles);
```
Spiral-Memory, line 19: Can we rather have useMemberStore here ? It is more related to member right ?

thesynthax: Yes, we can do this. Either way this would work

PR #528, opened 2024-03-22
fix: fixed static message limit to adapt to workspace settings

### LINKED ISSUE #527: [BUG] Static message Limit
### Description:
Currently, there is a static message limit in the embedded chat. However, this should not be the case as the admin has the flexibility to adjust the message limit according to their choice from workspace settings.

### Steps to reproduce:

1. Setup Embedded Chat.
2. Go to workspace settings and change the message limit.
3. Try sending a message from the embedded chat; it will always take 500 messages as the maximum limit.

### Expected behavior:

The maximum message limit should reflect the limit set by the admin.

### Actual behavior:

The maximum message limit is set to a constant value of 500 which is incorrect.

Spiral-Memory:
# Brief Title
This pull request is intended to fix the static message limit set in the code.

- [x] Added function to fetch the maxMessage limit setting.
- [x] Upon user login, fetch this setting and set it in the global state immediately.
- [x] Utilize that limit to determine whether to send as a message or attachments.

Fixes #527 

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/78961432/ca8898ee-1770-40ea-9827-9cbdd4854a1b

Note : This requires a rebuild of EC as there are some backend codes introduced.


Spiral-Memory:
Deleted by mistake !

PR #540, opened 2024-04-06
fix: centered throbber, added fallback, set default closable false

Spiral-Memory:
# Brief Title
Aligned the throbber in the image gallery and fixed infinite loading even when image fetching fails in the newly added image gallery (The same is implemented in RC as well)

Separately, in relation to another issue: `isClosable` is set to true in story examples, causing confusion. It should only be enabled when the true prop is explicitly passed along with a proper `setClosable` function, rather than just a boolean value (which was causing errors in the console). To avoid confusion, I have set `setClosable` as false by default and am passing an empty function.

- [x] Throbber centered in Image Gallery
- [x] Fallback text displayed when image fetch fails, preventing infinite loading
- [x] `isClosable` set to false by default in all stories
- [x] Default to an empty function for `setClosable` 

Fixes N.A

### Before:

image

image

https://github.com/RocketChat/EmbeddedChat/assets/78961432/1e11cd83-5b6b-4a6a-baff-115f725b8e93

### After:

image

image

https://github.com/RocketChat/EmbeddedChat/assets/78961432/54e605e7-3395-4c42-9878-467b2b8b75fa

### DIFF
--- a/packages/react/src/components/Attachments/ImageGallery.js
+++ b/packages/react/src/components/Attachments/ImageGallery.js
@@ -6,23 +6,30 @@ import classes from './ImageGallery.module.css';
 import { Throbber } from '../Throbber';
 import { ActionButton } from '../ActionButton';
 import { Icon } from '../Icon';
+import { Button } from '../Button';
 
 const ImageGallery = ({ currentFileId, setShowGallery }) => {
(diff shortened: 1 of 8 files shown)


sidmohanty11 (approved):
Thanks @Spiral-Memory : )

PR #557, opened 2024-04-10
feat: realtime sidebars

### LINKED ISSUE #543: [FEAT]: Adding real-time fetching of threads, mentions, etc. in sidebars.
As an EmbeddedChat developer,

**I need to**:
Listen to real-time APIs to update all threads, mentions, files, pinned messages, and starred messages (whichever is possible).

**So That**:
They are updated directly in the sidebar menu as soon as a new item is generated.

**Acceptance Criteria**:

- [ ] Ensure the list of items in the sidebar is updated in real-time.

Spiral-Memory:
# Brief Title
This PR adds real-time updates for pinned, starred, and mentioned messages in the sidebar by utilizing messages directly as messages are updated in realtime, rather than explicitly fetching them from APIs.

- [x] Real-time update integration in the pinned messages sidebar
- [x] Real-time update integration in the starred messages sidebar
- [x] Real-time update integration in the mentioned messages sidebar

Fixes #543 

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/78961432/0e8580ce-d018-4fbb-a501-858738dde9d7

### DIFF
--- a/packages/react/src/components/PinnedMessages/PinnedMessages.js
+++ b/packages/react/src/components/PinnedMessages/PinnedMessages.js
@@ -1,7 +1,6 @@
-import React, { useState, useContext, useEffect } from 'react';
+import React, { useState, useEffect } from 'react';
 import { isSameDay, format } from 'date-fns';
-import RCContext from '../../context/RCInstance';
-import { usePinnedMessageStore } from '../../store';
+import { usePinnedMessageStore, useMessageStore } from '../../store';
 import { Box } from '../Box';
 import { Icon } from '../Icon';
 import { MessageDivider } from '../Message/MessageDivider';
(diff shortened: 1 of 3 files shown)


Spiral-Memory:
@sidmohanty11 , I have used the messages array to extract all the details for pinned messages, starred messages, and mentioned messages, as the messages array is updated through the real-time API. In threads, it is already updated in real-time as it uses the messages array.

Regarding the user status of members, I couldn't find any real-time API that can handle this. For files, many details necessary to render the sidebar are not available in the "attachments" key. Hence, it's better to utilize the REST API instead of messages. Even in the Rocket.Chat app, the information is incomplete and requires a refresh to render the complete details of a file.

I have also eliminated unnecessary REST API calls. Please let me know if you would like me to keep them or if it would be better to remove them. If you prefer to keep them, I will revert my last commit.


Edit: Even in Rocket.Chat, it can't show the complete detail without a refresh (refer to the first attachment, where only the icon with no file info is shown), so I don't think it will be good to add real-time updates to files at this stage. When Rocket.Chat adds complete details of the file into messages itself, at that time we can consider it, or if at a later stage it provides a real-time API to update files, then it can be done.

image

Spiral-Memory:
@JeffreytheCoder  @sidmohanty11  Is it fine to use `useEffect` here with messages as the dependency array as it will only recalculate when messages array changes? Is there a way to further optimize it? If I use `useMemo` instead, how will I be able to handle the loading state? Please let me know your thoughts guys.

JeffreytheCoder:
I guess you can separate filtering message to `useMemo` and updating loading state to `useEffect`, both having `messages` as the dependency. ChatGPT says `useMemo` runs before the component re-renders and `useEffect` runs afterwards. You can try it out to see if it works.

Spiral-Memory:
I  tried that approach, but there wasn't much effect. Also, separating loading in useEffect won't work properly, as we want to stop loading once useMemo finishes its job. For the first loading, it will be fine using it, but for further loading, it won't work.

Okay then, let's keep it like this for now.

If I see any better approach for optimization, I'll do it separately!

@sidmohanty11 Ready for review.

sidmohanty11 (approved):

PR #598, opened 2024-06-26
Fix: read only channels to work the same for all users

### LINKED ISSUE #541: [bug]: Issues with Read Only channel options
### Description:
Currently the following are the issues of 'read-only' channel.
1. The read-only is not set dynamically to EC but is only set after page reload.
2. The read-only is also set for admins which should not be the case.
3. No appropriate message is displayed on EC when read-only option is toggled.

### Steps to reproduce:

1. Go to Rocket.Chat server
2. Set to channel to read-only and see the behaviour.

### Expected behavior:
The above mentioned issues should be resolved.

### Actual behavior:
[readonly.webm](https://github.com/RocketChat/EmbeddedChat/assets/95426993/b68ac887-3548-40f8-896d-5ac27a9af993)

devanshkansagra:
# Brief Title

- [ ] Task 1: The read-only is not set dynamically to EC but is only set after page reload.
- [x] Task 2: The read-only is also set for admins which should not be the case
- [ ] Task 3: No appropriate message is displayed on EC when read-only option is toggled.

Fixes #541

## Code that fixed
```(in ChatHeader.js => getChannelInfo() function)```
```.js 
      const userRes = await RCInstance.getAuth();
      const isAdmin = userRes.currentUser.me.roles.includes('admin');
      if (res.success) {
        setChannelInfo(res.room);
        if (res.room.t === 'p') setIsChannelPrivate(true);
        if (res.room.ro && !isAdmin) {
          setIsChannelReadOnly(true);
          setMessageAllowed();
        }
```

## Video/Screenshots
[Screencast from 2024-06-26 12-08-54.webm](https://github.com/RocketChat/EmbeddedChat/assets/125076549/6a9856f7-9036-4e6d-b878-af9a9772da3f)


packages/react/src/views/ChatHeader/ChatHeader.js
```diff
@@ -158,12 +160,15 @@ const ChatHeader = ({
       }
     };
 
+    
     const getChannelInfo = async () => {
       const res = await RCInstance.channelInfo();
+      const userRes = await RCInstance.getAuth();
+      const isAdmin = userRes.currentUser.me.roles.includes('admin');
```
Spiral-Memory, line 167: Even for admins, we want to display the 'This room is read-only' label. We should only enable the input box for admins in that case (which is disabled for all currently). Additionally, we can't just include admins; it depends on the workspace admin to allow people to message in read-only channels. If a role includes any permissions related to that, it should enable messaging

image

devanshkansagra: Did you mean this label?
image

Spiral-Memory: Yes

Spiral-Memory: Is it hidden in Rocket.chat main application for admins too ? Once check and let me know !

Spiral-Memory: Also, if you will see a variable : const allRoles = [...channelLevelRoles, ...workspaceLevelRoles]; Just use it and compare with PostReadOnly roles

Spiral-Memory:
Hey, I think you need not to do these changes. In the following code:

```javascript
const canSendMsg = postMsgRoles.some((role) => allRoles.includes(role));
```

Just remove `channelLevelRoles.length > 0 &&`, and everything will start working properly."

devanshkansagra:
image

Spiral-Memory:
Yep, now check if things are working fine !

Spiral-Memory (approved):

Spiral-Memory:
LGTM !

devanshkansagra:
image

sidmohanty11 (approved):
Thanks for the contribution @devanshkansagra!

PR #599, opened 2024-06-27
feat: configure EC remotely

### LINKED ISSUE #597: [FEAT]: Remotely configure EmbeddedChat props through RC app
As an EmbeddedChat developer,

**I need to**:
Implement functionality to remotely configure EmbeddedChat props

**So that**:
Users can customize the settings through their Rocket.Chat workspace

**Acceptance Criteria**

- [x] Fetch the props from EC settings and override the component props if present

Spiral-Memory:
# Brief Title
This PR adds a feature to remotely configure EmbeddedChat using the EmbeddedChat RC app.

- [x] Added all the required string/boolean/select settings in the EmbeddedChat RC app
- [x] Refactored the EmbeddedChat RC app's read setting logic
- [x] Introduced a fetch function to read RC app settings
- [x] Added a `useRemoteProps` hook that overrides the existing props if any of the props match those in the remote configuration

The implementation is simple: before displaying the EmbeddedChat, we fetch the remote props from EC settings. If a new prop is present or an existing one needs to be overridden, it overrides the prop and then shows the EmbeddedChat component. These overrides happen only when `remoteOpt` is passed as true.

We can't directly mutate the props, so I have added a state config that now handles the props accordingly.

Fixes #597

## Video/Screenshots

https://github.com/RocketChat/EmbeddedChat/assets/78961432/8f2e5027-2a8d-4723-8c2b-33ae7cbf9336


Spiral-Memory:
Hey @sidmohanty11 ,

This PR is ready for review. It fixes some minor issues and mainly adds a feature to remotely configure the props. I've explained its workings in the PR description. Please take a look.

Thanks!

packages/rc-app/settings/settings.ts
```diff
@@ -44,3 +44,165 @@ export const settings: ISetting[] = [
         packageValue: "",
     },
 ];
+
+export const propSettings: ISetting[] = [
```
sidmohanty11, line 48: question: can we add validation to each setting?

packages/rc-app/settings/settings.ts
```diff
@@ -44,3 +44,165 @@ export const settings: ISetting[] = [
...
+        required: false,
+        public: false,
+        packageValue: true,
+    },
+    {
+        id: "toast-bar-position",
+        i18nLabel: "Toast Bar Position",
+        i18nDescription:
+            "Specifies the position at which the toast bar is displayed.",
+        type: SettingType.STRING,
```
sidmohanty11, line 104: can we have a select for this?

Spiral-Memory: Yes sure !

packages/react/.eslintrc.json
```diff
@@ -45,6 +45,7 @@
     "import/no-cycle": 0,
     "import/no-extraneous-dependencies": "off",
     "react/no-unknown-property": ["error", { "ignore": ["css"] }],
-    "react/prop-types": "off"
+    "react/prop-types": "off",
+    "react/no-unused-prop-types": "off"
```
sidmohanty11, line 49: why was this required

packages/rc-app/settings/settings.ts
Spiral-Memory: Validation in the sense @sidmohanty11  ?

packages/react/src/views/EmbeddedChat.js
```diff
@@ -178,6 +201,10 @@ const EmbeddedChat = ({
     [RCInstance, ECOptions]
   );
 
+  if (!useRemoteProps(remoteOpt, RCInstance, setConfig)) {
```
sidmohanty11, line 204: we shouldn't use hooks like this

Spiral-Memory: Check now, if this looks better ?

packages/react/.eslintrc.json
Spiral-Memory: Since I am no longer directly using the props received in EmbeddedChat, but rather using them to initialize the config state, and then extracting the info from config state, the propTypes will throw a warning that props are not used. However, they are still being shown in propTypes.

packages/react/src/views/EmbeddedChat.js
```diff
@@ -82,18 +100,23 @@ const EmbeddedChat = ({
 
   const [RCInstance, setRCInstance] = useState(() => initializeRCInstance());
 
+  const reInstantiate = useCallback(() => {
```
sidmohanty11, line 103: why do we need to re-instantiate?

packages/rc-app/settings/settings.ts
sidmohanty11: for example, width values can only be CSS width values. Maybe we can zod validator or something similar to check for the correctness of the arbitrary string

packages/react/src/views/EmbeddedChat.js
Spiral-Memory: It was there earlier as well, but it was unstructured and caused unnecessary renders. I just fixed some linting issues, that's it. The reason why it is there is because, if someone changes the host, auth, or roomId from Storybook controls, it has to re-establish the connection with that roomId or a new host. If you look at the dependency array, it only includes host, roomId, and auth.

packages/react/.eslintrc.json
sidmohanty11: got it! thanks

packages/react/src/views/EmbeddedChat.js
sidmohanty11: i feel like the hook to be redundant, it can be simplified if we just add a useEffect and check if (remoteOpt == true) {  } then we would not need to pass the RCInstance or a setter to a hook

Spiral-Memory: Yep sure, that can be done, i just made it like this to reduce the amount of code in EmbeddedChat main component.. and keep it clean.. I'll change the structure.

packages/rc-app/settings/settings.ts
Spiral-Memory: We can't use external npm packages in RC app development for now. Most likely, it will not compile. I have tried integrating external packages while making RC app prototypes, and it throws errors. It is in the experimental phase for now.

https://developer.rocket.chat/docs/npm-package-support-experimental

packages/react/src/views/EmbeddedChat.js
Spiral-Memory: Done!

packages/rc-app/settings/settings.ts
Spiral-Memory: Done

packages/rc-app/settings/settings.ts
sidmohanty11: can we validate it by creating util functions on the app side? Not adding any extra package just some basic ones.

for example for width we can check if the value is a number and contains px or vw or % etc?

packages/react/src/views/EmbeddedChat.js
```diff
@@ -178,6 +222,8 @@ const EmbeddedChat = ({
     [RCInstance, ECOptions]
   );
 
+  if (!isSynced) return null;
```
sidmohanty11, line 225: i think we only need this if remoteOpt == true right?

sidmohanty11: i think this will always be false and lead to null if we have remoteOpt false?

Spiral-Memory: No, it won't. If `remoteOpt` is false in `useEffect`, it will return early. However, the finally block will still run and change `isSynced` to true.

packages/rc-app/settings/settings.ts
Spiral-Memory: I don't think we can able validation in rc app settings, it doesn't accept any rules or anything that can check this as far as i know, but I'll surely check once and ask someone if this is possible !!

packages/react/src/views/EmbeddedChat.js
sidmohanty11: IMO its bit unreadable, lets just refactor it out such that the function only gets called when/if remoteOpt is true else I don't think we need to be calling this function

sidmohanty11 (approved):
LGTM! Awesome work!

PR #600, opened 2024-06-29
Fixed unwanted close call in ChatBody.js

devanshkansagra:
# Removed the close statement in ChatBody.js

- Due to close statement in ChatBody.js the websocket connection gets disconnected when the page refresh or it runs in new tab

### DIFF
--- a/packages/react/src/views/ChatBody/ChatBody.js
+++ b/packages/react/src/views/ChatBody/ChatBody.js
@@ -166,7 +166,7 @@ const ChatBody = ({
       setScrollPosition(messageListRef.current.scrollTop);
       setIsUserScrolledUp(
         messageListRef.current.scrollTop + messageListRef.current.clientHeight <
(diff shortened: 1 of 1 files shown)


Spiral-Memory (approved):

Spiral-Memory:
LGTM. Can you please add one more change to this? Let's separate the `useEffect` which calls `getMessagesAndRoles` so that it doesn't affect the listeners.

Like this:

```javascript
useEffect(() => {
  RCInstance.auth.onAuthChange((user) => {
    if (user) {
      getMessagesAndRoles();
    } else {
      getMessagesAndRoles(anonymousMode);
    }
  });
}, [RCInstance, anonymousMode, getMessagesAndRoles]);
```

image

Spiral-Memory (approved):

Spiral-Memory (approved):

sidmohanty11 (approved):

PR #606, opened 2024-07-09
feat: improve pkg size while bundled

Spiral-Memory:
# Brief Title

This PR aims to reduce the size of `packages/react` when bundled with `ui-elements` and other components.

- [x] Added `preserveModules`.

Fixes N.A.

### DIFF
--- a/packages/markups/package.json
+++ b/packages/markups/package.json
@@ -58,7 +58,6 @@
         "rimraf": "^5.0.1",
         "rollup": "^2.70.1",
         "rollup-plugin-analyzer": "^4.0.0",
-        "rollup-plugin-bundle-size": "^1.0.3",
         "rollup-plugin-dts": "^6.1.1",
         "rollup-plugin-peer-deps-external": "^2.2.4",
         "rollup-plugin-postcss": "^4.0.2",
--- a/packages/markups/rollup.config.js
+++ b/packages/markups/rollup.config.js
@@ -4,13 +4,11 @@ import babel from '@rollup/plugin-babel';
 import postcss from 'rollup-plugin-postcss';
 import external from 'rollup-plugin-peer-deps-external';
 import json from '@rollup/plugin-json';
-import bundleSize from 'rollup-plugin-bundle-size';
 import { terser } from 'rollup-plugin-terser';
 import replace from '@rollup/plugin-replace';
 import analyze from 'rollup-plugin-analyzer';
 import dts from 'rollup-plugin-dts';
 
-const packageJson = require('./package.json');
 const PRODUCTION = process.env.NODE_ENV === 'production';
 
 export default [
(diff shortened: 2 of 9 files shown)


sidmohanty11 (commented):
let's not implement this change, as it will not be beneficial when developing different packages. we should instead keep the packages as dependencies only (as you previously added). this change would also break the single file format, meaning it cannot be used/imported with one script.

other drawbacks i can think of are:

- for any change made to our `ui-elements` package, we would need to redeploy our react app, even if no specific change has been made to it.
- the development and deployment procedures will become difficult to manage.

Spiral-Memory:
This change is mandatory, @sidmohanty11 .

I noticed one more issue: without `preserveModules` set to true, even if we keep it as an external dependency, the load time will increase, and the app is becoming slower than usual. Storybook also seems to be finding it difficult to build, and the resource size is increasing significantly. However, this change manages that properly, and the application will return to its previous speed and load times.

Regarding the issue you mentioned, there won't be a problem. Even after this structure change, we can import it directly from a single point because, along with `preserveModules`, `preserveModulesRoot` is also set to `src`.

This PR is ready for review and can be merged without any issues.

If we don't merge this, the application performance will degrade significantly.

Spiral-Memory:
You can also refer to [this article](https://blog.harveydelaney.com/creating-your-own-react-component-library/). It's a great article in which he explains how to build a component library with Rollup.

At the end, he also suggests performing code splitting in such libraries if the number of exports/imports is quite high. Otherwise, this will happen:

Other component libraries also do code splitting like this and do not have everything in a single `index.js` file. Without code splitting, it will not be able to perform tree shaking efficiently, resulting in large package sizes and resource issues.

Spiral-Memory:
We'll still keep it as external dependency.. but code splitting is absolutely necessary for such component library 🥲..

Spiral-Memory:
If preserveModules is not set to true: 





If we set preserveModules to true, it will return to it's original resource size which was there before implementing monorepo structure

sidmohanty11:
> We'll still keep it as external dependency.. but code splitting is absolutely necessary for such component library 🥲..

Ahh makes sense. my bad I thought we also updated rollup config of our React app to this format and thanks for clarifying that we are still keeping it as an external dependency (I thought we aren't). Looks good to me!

sidmohanty11 (approved):

Spiral-Memory:
> > We'll still keep it as external dependency.. but code splitting is absolutely necessary for such component library 🥲..
> 
> Ahh makes sense. my bad I thought we also updated rollup config of our React app to this format and thanks for clarifying that we are still keeping it as an external dependency (I thought we aren't). Looks good to me!

Got it! Just to clarify again, the EmbeddedChat package will still be the same as before. These changes have been applied to the component library and not to react. This will perform code splitting, helping with efficient imports and the removal of dead/eliminated code.

I'm also continuously learning about Rollup and builds and will include more updates to make it more optimized. I'm also new to this, so really really sorry for trouble 🥺❤️

Issue #629, opened 2024-09-29
FEAT: Make EmbeddedChat more mobile responsive


Spiral-Memory:
As an EmbeddedChat developer:

**I need to**:
Make EmbeddedChat more mobile-friendly, though it currently is.

**So that**:
It works well on mobile devices and in smaller-sized integrations.

**Acceptance Criteria**:

- [x] Identify potential issues on smaller screens.
- [x] Fix those issues to make the platform mobile-friendly.

Some known issues include:

1. On smaller screen sizes, the modal takes up the entire screen. It would be better to have some spacing, with the popups centered.
2. On smaller screens, the sidebar should open as a full-page view, covering the entire screen, with an option to go back.
3. Currently, the popups are strictly positioned. A more flexible positioning system is required. On mobile devices, these popups can act as modals or pages whichever suits better.
4. There may be additional responsive issues, and addressing as many of them as possible is appreciated.

Happy contributing!

Spiral-Memory:
Feel free to contribute to this issue. Your help is greatly appreciated.

yours-majesty:
@Spiral-Memory I really wanted to improve its UI too and improve the responsiveness a little more.

shrutika-gawande:
Hi @Spiral-Memory
I’d like to work on improving the mobile responsiveness of EmbeddedChat as described here.
I have experience with React, HTML, and CSS.
Could you please confirm if this issue is still open for contribution?

deepak0x:
@Spiral-Memory is the issue still open?
If yes will like to work on it...

Spiral-Memory:
Closing due to inactivity.

PR #634, opened 2024-10-01
Fixed UI issues for Sidebar

abirc8010:
# Brief Title
This PR fixes UI issues for Sidebar making it responsive

- [ ] Introduced a variable isSmallScreen to detect if the width of screen is <= 780px and adjusts the style of the component accordingly 
- [ ] For small screens , styled the Sidebar with position: absolute and covering entire width by setting width to 100%

Fixes # (issue)
#629
#685 

## Video/Screenshots
[Screencast from 2024-10-02 02-07-54.webm](https://github.com/user-attachments/assets/66f27643-c3e6-456f-828e-2439b61d42d0)

Popup Mode:

[popup.webm](https://github.com/user-attachments/assets/3a4d9282-6faa-44ed-bf93-ad4f9e9e03e8)

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-634 after approval.


abirc8010:
@Spiral-Memory do I need to test it on some more cases ?

Spiral-Memory:
Really loved the thorough testing done by you in the attached video, I'll review the PR and test myself once and let you know. 

Thank you so much for your contribution ❤️

Spiral-Memory:
Once go to design a variant, modern design and check if popup mode is working fine after this change and attach a video

abirc8010:
I've attached the popup mode video , 
I haven't made any changes for the popup mode for now though ,  I introduced a conditional statement so that the styles are applied for the Sidebar component only

abirc8010:
Hey @Spiral-Memory , are any more tests required ?

Spiral-Memory:
I haven't tested it yet, let the pr preview auto deployment feature get implemented, afterwards, I'll test and let you know if any changes are required

Spiral-Memory (approved):

Spiral-Memory:
Hey @abirc8010,

I noticed two issues. Please fix them:

1. In normal mode, the sidebar title is properly visible, but when switching to a smaller screen (with less height), the title is not visible. Ensure that in both modes, the title aligns at the same height.

image
image

2. While in dev mode, the switch does not happen automatically when changing screen sizes. I have to manually open and close the sidebar to reflect the changes, which is not the intended behavior.

abirc8010:
> 2. While in dev mode, the switch does not happen automatically when changing screen sizes. I have to manually open and close the sidebar to reflect the changes, which is not the intended behavior.

To fix this issue, I have to add resize event listener , will that be ok ?

devanshkansagra (approved):

Spiral-Memory:
> > 2. While in dev mode, the switch does not happen automatically when changing screen sizes. I have to manually open and close the sidebar to reflect the changes, which is not the intended behavior.
> 
> To fix this issue, I have to add resize event listener , will that be ok ?

But why do we need to listeners ? Why can't we use media queries to handle it properly ?

abirc8010:
> But why do we need listeners ? Why can't we use media queries to handle it properly ?

The Sidebar component accepts a style prop in the form of a css object, which makes it difficult to implement media queries effectively. That's why I am considering using a resize event listener instead .
Do you have any suggestions?

Spiral-Memory:
> > But why do we need listeners ? Why can't we use media queries to handle it properly ?
> 
> The Sidebar component accepts a style prop in the form of a css object, which makes it difficult to implement media queries effectively. That's why I am considering using a resize event listener instead . Do you have any suggestions?

It is not recommended to use event listeners in this use case. Try using only media queries. You can still write the CSS in the style.js file and pass it. Try to figure out a solution using media queries.

abirc8010:
@Spiral-Memory , I have fixed the issues

Spiral-Memory (commented):
The approach seems fine; you are trying to add it to the global styles. However, why can't we add similar media queries and pass them directly to the sidebars, without overriding class styles globally?

abirc8010:
I noticed that the Sidebar component from the embeddedChat ui-element only accepts style prop as JavaScript objects, which doesn't allow for media queries. As a result, I had to apply the styles using the classname in global styles .

image

Spiral-Memory:
As far as I'm aware, you can set media queries in a style object as well, or try using the css attribute with Emotion/React. I'll have to look into it to confirm.

Spiral-Memory (approved):
Approving to test !

Spiral-Memory:
@devanshkansagra 

Could you please check, if we can use media queries in style object ?.

abirc8010:
Fixed

Spiral-Memory (approved):
Approving to test !

abirc8010:
> Approving to test ! 

@Spiral-Memory do I need to make any changes?

Spiral-Memory:
I have tested this myself, and it looks good to me. @devanshkansagra  @smritidoneria , could you please review and test this PR as well at the following link? Let me know if you have any feedback or concerns: https://rocketchat.github.io/EmbeddedChat/pulls/pr-634

smritidoneria:
hey @Spiral-Memory , @abirc8010 , I noticed one issue

I noticed an issue with the thread messages sidebar. When we open the sidebar and click on "Reply to Thread," the user should be redirected to the thread-related chat section. However, this redirection isn’t happening as expected.

It works correctly on a tablet viewport but fails on larger and smaller mobile viewports.
https://github.com/user-attachments/assets/b509218b-0b15-429a-b4ef-004f2904a252

Spiral-Memory:
Nice catch, @smritidoneria ! Really appreciate it. @abirc8010, please take a look.

I have one concern: if anyone could test this, it would be great. Since it's a global css change, I'm a bit unsure how it will behave when embedded in a website

@smritidoneria , if you know how to embed the chat in a website and test it, please try in the right corner with height set to 70% and width set to 40% or something around that.

Otherwise, @devanshkansagra , please handle the testing.

abirc8010:
Hey @Spiral-Memory I fixed the issue , now the sidebar will close on clicking reply in thread

Spiral-Memory (approved):

Spiral-Memory:
@devanshkansagra 

Can you please test this after integrating it to some page and let me know if things are working fine

Thanks

devanshkansagra:
hey @Spiral-Memory , I am not able to publish it to npm to test, and also the official npm package of embeddedChat is not yet updated

Spiral-Memory:
Ok @devanshkansagra 
No worries.. I'll publish an unofficial package and test

packages/react/src/views/Message/Message.js
```diff
@@ -1,6 +1,6 @@
 import React, { memo, useContext } from 'react';
 import PropTypes from 'prop-types';
-import { format } from 'date-fns';
+import { format, set } from 'date-fns';
```
Spiral-Memory, line 3: importing set ?

packages/react/src/views/Message/Message.styles.js
```diff
@@ -75,7 +75,7 @@ export const getMessageDividerStyles = (theme) => {
       line-height: 1rem;
       position: relative;
       display: flex;
-      z-index: 1000;
+      z-index: 1;
```
Spiral-Memory, line 78: I'm a little afraid about z-index changes. It might disturb the existing flow in some situation. Is this required?

abirc8010: image

@Spiral-Memory , actually due to a high value of z-index, the message divider was appearing on top of the sidebar like the image shown , so I had to decrease the value.

packages/react/src/views/Message/Message.js
abirc8010: I think it was automatically imported by vs code . It doesn't have a use case in this context, I'll remove it.

Spiral-Memory:
Hi @abirc8010 

I really want to merge this PR. Could you please address the concerns I raised regarding the `z-index` and the unused `set`?  

I understand that the `z-index` is necessary, but could you modify it so that the new `z-index` is applied only at a specific breakpoint? When I initially set it up, I followed a specific set of `z-index` values to maintain consistency. That's why I can't merge it with a `z-index` of `1` across all screens.

abirc8010:
Hey @Spiral-Memory I reverted the zIndex for all except for the message divider , I used media query to reduce the zIndex for lower screen size to prevent it from appearing over the sidebar

packages/react/src/views/MessageAggregators/common/MessageAggregator.js
```diff
@@ -70,7 +70,7 @@ export const MessageAggregator = ({
       iconName={iconName}
       searchProps={searchProps}
       onClose={() => setExclusiveState(null)}
-      style={{ padding: 0 }}
+      style={{ zIndex: 1 }}
```
Spiral-Memory, line 73: These are still present

packages/react/src/views/Message/Message.js
Spiral-Memory: remove this too

abirc8010:
@Spiral-Memory I have reverted the zIndexes which were already present , can I add zIndex for the ViewComponent for smaller screens ? 
Some of the areas are getting disturbed if I don't adjust the zIndex

Spiral-Memory:
For smaller screen, you can specifically add zIndex according to the need.. i just don't want to let zindex changes to have affect in normal view

abirc8010:
Hey @Spiral-Memory I have made the required changes

Spiral-Memory (approved):

packages/react/src/views/GlobalStyles.js
```diff
@@ -36,6 +35,20 @@ const getGlobalStyles = (theme) => css`
   .ec-embedded-chat ::-webkit-scrollbar-button {
     display: none;
   }
+  .ec-sidebar {
```
Spiral-Memory, line 38: Why in global styles ? Can't we add it in specific sidebar component itself ?

Spiral-Memory (approved):

Spiral-Memory:
Looks good to me. Thanks a lot, @abirc8010  

Please add some good padding to the sidebars as well, notice that it is very less between the avatar and the side corner but you can include that in another PR. Since this has been pending for a long time, I am merging it.

image

PR #654, opened 2024-11-02
Fix UI Not Updating Immediately After Pinning/Unpinning Messages

### LINKED ISSUE #653: UI Not Updating Immediately After Pinning/Unpinning Messages
### Description:

When pinning or unpinning a message in the chat, the UI does not immediately reflect the change. The pin icon remains unchanged until the page is refreshed. Additionally, the system allows pinning the same message multiple times without updating the UI to show the message is already pinned.

### Steps to reproduce:

1. Open the Chat
2.Pin a message.

### Expected behavior:

The pin icon should change immediately after pinning or unpinning a message.
The system should not allow pinning a message that is already pinned without updating the UI.

### Actual behavior:

https://github.com/user-attachments/assets/b1914e14-7e13-454e-bb16-aa067c096795

smritidoneria:
# Brief Title
This PR addresses the issue where the UI does not immediately reflect the change when pinning or unpinning a message. The pin icon remains unchanged until the page is refreshed, and the system allows pinning the same message multiple times without updating the UI to show the message is already pinned.

- Optimistically update the UI by changing the pinned status of the message immediately before making the asynchronous call to pin/unpin the message.
- Revert the change if the asynchronous call fails.

Fixes #653

## Video/Screenshots

https://github.com/user-attachments/assets/fc985902-c59a-4c12-8167-95d52b34f154

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-<pr_number> after approval.

### DIFF
--- a/packages/react/src/views/Message/Message.js
+++ b/packages/react/src/views/Message/Message.js
@@ -91,10 +91,13 @@ const Message = ({
 
   const handlePinMessage = async (msg) => {
     const isPinned = msg.pinned;
+    // Optimistically update the UI
+    msg.pinned = !isPinned;
     const pinOrUnpin = isPinned
       ? await RCInstance.unpinMessage(msg._id)
       : await RCInstance.pinMessage(msg._id);
     if (pinOrUnpin.error) {
+      msg.pinned = isPinned;
       dispatchToastMessage({
         type: 'error',
         message: 'Error pinning message',


packages/react/src/views/Message/Message.js
```diff
@@ -91,10 +91,13 @@ const Message = ({
 
   const handlePinMessage = async (msg) => {
     const isPinned = msg.pinned;
+    // Optimistically update the UI
```
Spiral-Memory, line 94: Please remove this comment

Spiral-Memory (approved):

Barrylimarti:
Hey @smritidoneria! Although I understood your solution, but I am unable to figure out why this needs to be done manually. I think the best way would be if state updation triggers this action rather than setting this up optimistically. Do you know why the automatic updation is failing?

smritidoneria:
hey @Barrylimarti , previously the problem was that the code is directly modifies the msg.pinned property without triggering a re-render of the component. In React, directly mutating an object does not cause the component to re-render. React only re-renders components when the state or props change. Re-rendering was only happening at refreshing. that is why the previous code was failing.

so the solution for this comes out to be either use react states to manage the pinning of the message or optimistically update the ui. In my judgement, 2nd option was good to look forward to.

Spiral-Memory:
Yeah that's correct ! But how exactly does this new code work ? I think here also, you're updating msg.pinned property directly here as well.

Barrylimarti:
Okay, but if you see the case of starring the message, it does not require this manual update though. Isn't it strange? If there's a call to the starring api it triggers the re-render but not in the case of the pinning api?

smritidoneria:
Hey @Barrylimarti , Sorry for the late response. Actually , The handleStarMessage function does not directly update the msg.starred property.  Instead, it relies on the asynchronous operation RCInstance.starMessage or RCInstance.unstarMessage to update the state. This means that the state management for starred messages is handled elsewhere, likely in a parent component or a global state management solution (e.g., Redux, Context API).

But the handlePinMessage  function directly reads the msg.pinned property but does not update it directly.

Barrylimarti:
@smritidoneria Yes, that's exactly my point. But unpinning works well right? That means the same state management operations that is there for starred messages is there for pinning also but there's some bug while pinning it not while unpinning it. Hence, I think we should try to find that bug first at the state management level.

> But the handlePinMessage function directly reads the msg.pinned property but does not update it directly.

Also, I don't think this is the case as unpinning means updating the msg.pinned property to false, which is anyways happening.

Spiral-Memory:
@smritidoneria @Barrylimarti 

Did you guys figured out, what's the actual reason behind this ?

smritidoneria:
yes, it figured it out.
In the starred message feature, at the end of the process, the getStarredMessages() function is called. In the component utilizing useFetchChatData, we maintain a state for starred messages. When a user stars a message, the getStarredMessages() function triggers, which updates the state of the starred messages. This state change causes the component to re-render, enabling real-time updates in the starred message modal.
However, this is not the case with pinned messages. We are not maintaining any state for pinned messages, so there is no state change to trigger a re-render. As a result, real-time updates are not reflected in the pinned message modal.

smritidoneria:
We now have two possible approaches to address this issue:
1.State Management for Pinned Messages: Similar to starred messages, we can implement state management for pinned messages. This would allow state changes to trigger re-renders and enable real-time updates in the pinned message modal.
2.Proposed Solution in this PR

Spiral-Memory:
Let's go with this change for now

PR #657, opened 2024-11-04
Trim Whitespaces in Username/Email Fields in EmbeddedChat Login Form

### LINKED ISSUE #656: Whitespace in Username/Email Input Prevents Login in EmbeddedChat Login Form
### Description:
When logging in through the EmbeddedChat form, if extra whitespaces are added to the beginning or end of the username or email, the system doesn’t trim these whitespaces. This causes the login to fail, even though the credentials are otherwise correct.

### Steps to reproduce:

1. Go to the EmbeddedChat login form.
2. Enter a valid username or email, but include leading or trailing whitespaces.
3. Attempt to log in.

https://github.com/user-attachments/assets/1b9af213-686f-49b7-8945-a04624846140

From the above video it is clear that password's allow whitespaces along with text but the same is not allowed with username or email
so i shall not trim the password input box

### Expected behavior:
The system should automatically trim any whitespaces from the beginning or end of the username or email, allowing the login attempt to proceed if the credentials are correct.

### Actual behavior:
The system does not trim whitespaces, causing a login failure.

https://github.com/user-attachments/assets/1b47c6c1-3c7f-4dec-a5ce-515e0659530f

### Reason for Trimming Whitespaces:
- **Prevents Accidental Login Failures**: Users often accidentally add spaces before or after their input, especially on mobile devices. Trimming removes these unintended spaces, reducing the chance of failed login attempts due to formatting issues.
- **Aligns with User Expectations**: Most applications already trim whitespaces, so users typically expect that accidental spaces won’t affect their login.

SinghaAnirban005:
# Brief Title
Trim Whitespaces in Username/Email Fields in EmbeddedChat Login Form

- [x] Task 1: Add logic to trim whitespaces in username/email fields on login form submission.
- [x] Task 2: Test and verify that users can log in successfully even with accidental spaces.

Fixes #656 

## Video/Screenshots

https://github.com/user-attachments/assets/5303abd9-294c-4b1f-a697-5d0b8841eaba

I did not trim for passwords because while registering on rocket chat , they allow having whitespaces after some `password` has been written

## PR Test Details
**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-<pr_number> after approval.

### DIFF
--- a/packages/react/src/views/LoginForm/LoginForm.js
+++ b/packages/react/src/views/LoginForm/LoginForm.js
@@ -52,7 +52,7 @@ export default function LoginForm() {
   };
 
   const handleEdituserOrEmail = (e) => {
-    setUserOrEmail(e.target.value);
+    setUserOrEmail(e.target.value.trim());
   };
   const handleEditPassword = (e) => {
     setPassword(e.target.value);
--- a/packages/react/src/views/LoginForm/LoginForm.js
+++ b/packages/react/src/views/LoginForm/LoginForm.js
@@ -52,7 +52,7 @@ export default function LoginForm() {
   };
 
   const handleEdituserOrEmail = (e) => {
-    setUserOrEmail(e.target.value.trim());
+    setUserOrEmail(e.target.value);
   };
   const handleEditPassword = (e) => {
     setPassword(e.target.value);
--- a/packages/api/src/EmbeddedChatApi.ts
+++ b/packages/api/src/EmbeddedChatApi.ts
@@ -121,12 +121,12 @@ export default class EmbeddedChatApi {
     let credentials;
     if (!code) {
       credentials = credentials = {
-        user: userOrEmail,
+        user: userOrEmail.trim(),
         password,
       };
     } else {
       credentials = {
-        user: userOrEmail,
+        user: userOrEmail.trim(),
         password,
         code,
       };


Spiral-Memory:
Here's a corrected version:

@sidmohanty11 , what do you think? Is this necessary?

Spiral-Memory:
@sidmohanty11 , what do you think? Is this necessary?

devanshkansagra:
Yeah I also agree with @Spiral-Memory, why would someone add whitespaces in username/email?

Spiral-Memory:
He might be considering that, in case of a mistake while typing emails, any extra whitespace would be trimmed automatically, as some websites implement such features. However, I need to confirm this with Sidharth, since it's not strictly necessary, and there could be issues with this feature

sidmohanty11:
I think this is a good addition

packages/react/src/views/LoginForm/LoginForm.js
```diff
@@ -52,7 +52,7 @@ export default function LoginForm() {
   };
 
   const handleEdituserOrEmail = (e) => {
-    setUserOrEmail(e.target.value);
+    setUserOrEmail(e.target.value.trim());
```
Spiral-Memory, line 55: @SinghaAnirban005  When we set the credential object and use the .trim() function there, why do we need it here as well?

SinghaAnirban005: You’re absolutely right since I’ve already used .trim() when setting the credentials object, additional trimming is unnecessary. Thanks for the catch !!

Spiral-Memory (approved):

Spiral-Memory:
LGTM! 
Thanks for the addition @abirc8010

PR #674, opened 2024-12-01
Fix:  Restrict Pin Icon Visibility Based on User Permissions

### LINKED ISSUE #672: Non-admin users can see the pin/unpin icon in group chats despite lacking permissions
### Description:
Admins are authorized to pin/unpin messages and allow other non-priviledged users to pin/unpin based on their choice. However, the application currently displays the pin/unpin icon to non-priviledged or non-admin users, which is misleading and incorrect as they do not have the required permissions.

### Steps to reproduce:

1. Open any chat
2. Ensure you are not signed in as admin (you should be a non-admin user)
3. Try pinning any message using the pin icon

### Expected behavior:
Non-priviledged users should not see the pin/unpin icon in the group chat, as they are unauthorized to perform this action. Only admins and those user group chosen by the admin should have access to this functionality.

### Actual behavior:
The pin/unpin icon is visible to non-priviledged users in the group chat. Clicking the icon shows an error, leading to a confusing user experience.

https://github.com/user-attachments/assets/190912fd-8d2c-4cab-9bb7-e4b6b8a15daf

Below is the response generated on trying to pin/unpin a message as a non-admin
Screenshot 2024-11-29 000103

SinghaAnirban005:
# Brief Title

- [x]  Ensure users without permissions do not see the pin icon
- [x]  Verify pin icon visibility for users with admin-granted permissions.

Fixes #672 

## Video/Screenshots

Case 1:  When non-priviledged users are granted permission to pin/unpin messages

https://github.com/user-attachments/assets/9ed97726-bde6-4bc2-9686-3f2193bddd2d

Case 2: When non-priviledged users are not granted permission to pin/unpin messages

https://github.com/user-attachments/assets/aeb0f199-77da-4aa0-ba9b-56e91ed4262c

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-674 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.

### DIFF
--- a/packages/react/src/hooks/useRCAuth.js
+++ b/packages/react/src/hooks/useRCAuth.js
@@ -20,11 +20,13 @@ export const useRCAuth = () => {
   );
   const setPassword = useUserStore((state) => state.setPassword);
   const setEmailorUser = useUserStore((state) => state.setEmailorUser);
+  const setUserPermissions = useUserStore((state) => state.setUserPermissions);
   const dispatchToastMessage = useToastBarDispatch();
 
   const handleLogin = async (userOrEmail, password, code) => {
     try {
       const res = await RCInstance.login(userOrEmail, password, code);
+      const permissions = await RCInstance.permissionInfo();
       if (res.error === 'Unauthorized' || res.error === 403) {
         dispatchToastMessage({
           type: 'error',
@@ -56,6 +58,7 @@ export const useRCAuth = () => {
           setIsTotpModalOpen(false);
           setEmailorUser(null);
           setPassword(null);
+          setUserPermissions(permissions);
           dispatchToastMessage({
             type: 'success',
             message: 'Successfully logged in',
--- a/packages/react/src/hooks/useRCAuth.js
+++ b/packages/react/src/hooks/useRCAuth.js
@@ -20,7 +20,9 @@ export const useRCAuth = () => {
   );
   const setPassword = useUserStore((state) => state.setPassword);
   const setEmailorUser = useUserStore((state) => state.setEmailorUser);
-  const setUserPermissions = useUserStore((state) => state.setUserPermissions);
+  const setUserPinPermissions = useUserStore(
+    (state) => state.setUserPinPermissions
+  );
   const dispatchToastMessage = useToastBarDispatch();
 
   const handleLogin = async (userOrEmail, password, code) => {
@@ -58,7 +60,7 @@ export const useRCAuth = () => {
           setIsTotpModalOpen(false);
           setEmailorUser(null);
           setPassword(null);
-          setUserPermissions(permissions);
+          setUserPinPermissions(permissions.update[150]);
           dispatchToastMessage({
             type: 'success',
             message: 'Successfully logged in',
--- a/packages/react/src/views/Message/MessageToolbox.js
+++ b/packages/react/src/views/Message/MessageToolbox.js
@@ -21,6 +21,7 @@ export const MessageToolbox = ({
   style = {},
   isThreadMessage = false,
   authenticatedUserId,
+  isAllowedToPin,
   handleOpenThread,
   handleEmojiClick,
   handlePinMessage,
@@ -110,7 +111,7 @@ export const MessageToolbox = ({
         id: 'pin',
         onClick: () => handlePinMessage(message),
         iconName: message.pinned ? 'pin-filled' : 'pin',
-        visible: !isThreadMessage,
+        visible: !isThreadMessage && isAllowedToPin,
       },
       edit: {
         label: 'Edit',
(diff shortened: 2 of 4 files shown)


Spiral-Memory (approved):
LGTM

devanshkansagra:
Hey @SinghaAnirban005 your code is breaking for admin user, check this
image

Spiral-Memory:
Check if it's due to alphabet case?

devanshkansagra:
meaning?

Spiral-Memory:
lower/upper case difference while comparing

packages/react/src/hooks/useRCAuth.js
```diff
@@ -56,6 +60,7 @@ export const useRCAuth = () => {
           setIsTotpModalOpen(false);
           setEmailorUser(null);
           setPassword(null);
+          setUserPinPermissions(permissions.update[150]);
```
devanshkansagra, line 63: Hey @SinghaAnirban005, I think it is permissions.update[146] according to the previous pr - #520

devanshkansagra:
> lower/upper case difference while comparing

we are setting the permissions for pin messages according to the user roles, so how this lower/upper case difference, could you please explain?

packages/react/src/views/Message/MessageToolbox.js
```diff
@@ -67,6 +69,7 @@ export const MessageToolbox = ({
     setShowDeleteModal(false);
   };
 
+  const isAllowedToPin = userRoles.some((role) => pinRoles.has(role));
```
Spiral-Memory, line 72: @devanshkansagra 

At this point, we are comparing two different rules. There is a possibility that one of the rules is in a different case.

It's just a possibility, but I wanted to check.
I'm not sure if that's the issue.

The problem might lie in fetching the wrong PIN permission list,
or it could be that the fetch operation somehow fails.

devanshkansagra: Okay so you mean, the roles in the roles array?

Spiral-Memory: Suppose in userRole, my role is Admin 
In roles array, in pin permission, the role is admin 


Now Admin != admin
Hence check fails

devanshkansagra: Hey @Spiral-Memory when I was researching for the solution of this issue, I somehow fetched the entire roles array from the permissions response and in that I have seen all the roles were having same case all small

Spiral-Memory: Ohh ok then, 
Try figuring out the issue and let me know

devanshkansagra: Okay

packages/react/src/hooks/useRCAuth.js
SinghaAnirban005: The 150th index returns the pin permissions correctly

devanshkansagra: okay then what is on 146th??

devanshkansagra: check and try pr 520

packages/react/src/views/Message/MessageToolbox.js
SinghaAnirban005: @devanshkansagra @Spiral-Memory 
In my POV , the code seems to work well for all user groups including admin , the reason why it might not have worked is that when pin roles are changed by admin in the panel, changes are instantaneously not updated in the user's permission store . Thus a user needs to logout and re-login in order to see changes in permissions

packages/react/src/hooks/useRCAuth.js
SinghaAnirban005: i am getting this log on 146 
Screenshot 2024-12-17 004227

devanshkansagra: Ohh so might this permissions array got changed later

packages/react/src/views/Message/MessageToolbox.js
devanshkansagra: rather than doing logout and login again is it possible to view changes on refreshing the page?

packages/react/src/hooks/useRCAuth.js
SinghaAnirban005: Yes exactly

packages/react/src/views/Message/MessageToolbox.js
SinghaAnirban005: i can try fixing it !!

PR #777, opened 2025-01-02
Fixed the errors occured due to statically setting the values of various permissions

devanshkansagra:
I have observed and encountered these errors like `Cannot read the properties of undefined reading('150') or ('28') or ('70')` on the integrating the EmbeddedChat to a react application. Due to these errors the EmbeddedChat is not properly rendering on integration side. One demo on the integration of embeddedchat is `e2e-react` you can see these error over there also if you haven't integrated yet to any of the website.

## Video/Screenshots
Screenshot from 2025-01-02 15-44-18
Screenshot from 2025-01-02 15-45-34
Screenshot from 2025-01-02 15-48-39
Screenshot from 2025-01-02 15-52-30
Screenshot from 2025-01-02 15-53-14
Screenshot from 2025-01-02 15-54-18

As there will be new release of EmbeddedChat very very soon and till then there shouldn't be any broken output so I have removed that buggy code. Contributors are requested to come with dynamic and optimal approach when solving these kind of issues and also should use the latest stable release of RocketChat server.

Thank you

### DIFF
--- a/packages/react/src/hooks/useFetchChatData.js
+++ b/packages/react/src/hooks/useFetchChatData.js
@@ -20,9 +20,6 @@ const useFetchChatData = (showRoles) => {
   const isUserAuthenticated = useUserStore(
     (state) => state.isUserAuthenticated
   );
-  const setViewUserInfoPermissions = useUserStore(
-    (state) => state.setViewUserInfoPermissions
-  );
 
   const getMessagesAndRoles = useCallback(
     async (anonymousMode) => {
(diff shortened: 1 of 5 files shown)


Spiral-Memory:
Hi @devanshkansagra ,
Thanks for your effort, but this is not a valid solution to the issue you mentioned. You might want to investigate the root cause of the problem. The major reason I can think of is that React states are being set too late, which causes the code to attempt reading values from indexes that haven’t been set yet, resulting in this error.

I'm closing this PR for now. Once you figure out the exact cause, feel free to submit another PR.

PR #859, opened 2025-01-09
Feat: Added 'Mentions you' and 'Mentions user' tooltips for mentioned users & underline on hover

### LINKED ISSUE #858: Feat: Add 'Mentions you' and 'Mentions user' tooltips for mentioned users & underline on hover
As an EmbeddedChat developer

**I need to**: Add a feature that handles user mentions in the chat. When a user is mentioned, show a tooltip with a message. If the mention is for the logged-in user, show "Mentions you" in the tooltip. Show "Mentions user" in the tooltip if it's another user.

**So That**:  The UI will be consistent for all tooltips, and users will have a clear and accessible interaction with these elements.

**Acceptance Criteria**

- [ ] Show correct tooltip: If a mention is for the logged-in user, show the tooltip with the text "Mentions you". For other users, show "Mentions user".
- [ ] Styling: On hover, an underline should appear below the text.

dhairyashiil:
# Brief Title
Feat: Added 'Mentions you' and 'Mentions user' tooltips for mentioned users & underline on hover

- [X] Show correct tooltip: If a mention is for the logged-in user, show the tooltip with the text "Mentions you". For other users, show "Mentions user".
- [X] Styling: On hover, an underline should appear below the text.

Fixes #858

## Video/Screenshots

https://github.com/user-attachments/assets/6b7d0ecf-516d-4113-a668-f3bd975b4de3

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-859 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.


Spiral-Memory:
There are other tooltips too for @all and @here something like this.. Make sure you account those too.. if we are adding it.. though this doesn't look much necessary

dhairyashiil:
> There are other tooltips too for @ALL and @here something like this.. Make sure you account those too.. if we are adding it.. though this doesn't look much necessary

Thank you, @Spiral-Memory, for the review. I added tooltips for the remaining two cases. I somewhat agree with you that it doesn't seem entirely necessary, but for the case of "Mentions you," I think it is important.

For example, suppose there are two people in the room with similar usernames like 'dhairi12' and 'dhairl12'. 
The only difference is the characters "i" and "l" before the number 1. 
Without the tooltip, the user would have to constantly check the chat to determine whether they were mentioned or if it was the other user. They would need to check the spelling or navigate to the mentions section in the sidebar. 
But with this tooltip feature, they only need to hover over it.

dhairyashiil:
UI after recent commit: 

https://github.com/user-attachments/assets/5dc0ee58-8e18-4f1d-a660-876c7f937feb

Spiral-Memory:
> > There are other tooltips too for @ALL and @here something like this.. Make sure you account those too.. if we are adding it.. though this doesn't look much necessary
> 
> Thank you, @Spiral-Memory, for the review. I added tooltips for the remaining two cases. I somewhat agree with you that it doesn't seem entirely necessary, but for the case of "Mentions you," I think it is important.
> 
> For example, suppose there are two people in the room with similar usernames like 'dhairi12' and 'dhairl12'. The only difference is the characters "i" and "l" before the number 1. Without the tooltip, the user would have to constantly check the chat to determine whether they were mentioned or if it was the other user. They would need to check the spelling or navigate to the mentions section in the sidebar. But with this tooltip feature, they only need to hover over it.

Not exactly @dhairyashiil 
It is already differenciated with color

PR #901, opened 2025-01-15
Fix: Tool Tips positioned above for Message Toolbox

### LINKED ISSUE #900: Bug: Tooltip Content Not Fully Visible in UI Element
### Description:

The tooltips in the admin panel UI does not display fully, making it difficult to read the complete text. This issue affects usability and clarity for users.

### Steps to reproduce:

1. Hover over the icons in the message toolboox.
2. Observe that the tooltip content is cut off and not fully visible.

### Expected behavior:

The tooltip should be fully visible and display the complete text without being truncated.

### Actual behavior:

The tooltip is partially cut off and does not display the full text.

https://github.com/user-attachments/assets/5b077a80-4c95-4f71-8b24-013697ad3463

KrishnaShuk:
# Tool-Tips positioned above Message Toolbox

- [ ] The tooltip should be fully visible and display the complete text without being truncated.

Fixes #900 

## Video/Screenshots

[Screencast from 2025-01-15 15-20-10.webm](https://github.com/user-attachments/assets/a6c85ec3-a3fd-431e-a3c8-704419b4adea)

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr- after approval.

### DIFF
--- a/packages/react/src/views/Message/MessageToolbox.js
+++ b/packages/react/src/views/Message/MessageToolbox.js
@@ -229,9 +229,6 @@ export const MessageToolbox = ({
           style={styleOverrides}
           {...props}
         >
-          {surfaceOptions?.length > 0 && (
-            <SurfaceMenu options={surfaceOptions} size="small" />
-          )}
           {menuOptions?.length > 0 && (
             <Menu
               size="small"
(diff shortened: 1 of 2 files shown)


dhairyashiil:
Hello @KrishnaShuk, the tooltip arrow should point to the message toolbox, not the icon inside. If you want to change the position from bottom to top, please consider this point as well: Observe the current tooltip position, it's not directly on the icon.

KrishnaShuk:
Will lift up the tooltip arrow that will point to the message toolbox. @dhairyashiil thanks for suggestion.

dhairyashiil:
> Will lift up the tooltip arrow that will point to the message toolbox. @dhairyashiil thanks for suggestion.

Well, I have an idea. Instead of changing the position of the tooltip from bottom to top, you can change the positioning of the message toolbox by shifting it upwards. However, there's a catch: the message toolbar positioning differs for normal messages and quoted messages, so you need to ensure that the toolbar is usable for both types of messages.

Spiral-Memory:
If you shift it to top, then the problem will again be the same when a message is at the top i believe, can we make it dynamic ?

KrishnaShuk:
You are right @Spiral-Memory , will try to make it more dynamic.

PR #955, opened 2025-02-03
fix: No Loading Feedback on Login Form, Causing Unresponsiveness

### LINKED ISSUE #954: No Loading Feedback on Login Form, Causing Unresponsiveness
### Description:

There is no loading indication in the login form, making it difficult for users with slow internet connections to determine if the login button was clicked successfully. This makes the form feel unresponsive or broken. 

### Steps to reproduce:
1. Open Developer Tools (F12).
2. Go to the Network tab and enable 3G throttling.
3. Click the "Join" button.
4. Fill in the form and click "Login."
5. Observe that the form provides no feedback while the request is pending.

### Expected behavior:
The form should provide visual feedback (e.g. the `Throbber` loader) while the request is processing.

### Actual behavior:
The form provides no feedback, making it appear broken or unresponsive.
Video:
https://github.com/user-attachments/assets/db49a524-d9e2-4d36-b5fb-ddf2fe01a55d

Abbas-Askari:
# Add loader to login form

`Throbber` is added to the login form button while the request is pending and is removed afterwards, making the form more responsive.

Fixes #954 

## Video
VIdeo:
[Screencast from 2025-02-03 18-35-47.webm](https://github.com/user-attachments/assets/5e6a35db-146c-426f-9318-90c93d1ba4c4)

## PR Test Details
**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-955 after approval.

### DIFF
--- a/packages/react/src/views/LoginForm/LoginForm.js
+++ b/packages/react/src/views/LoginForm/LoginForm.js
@@ -7,6 +7,7 @@ import {
   Input,
   Icon,
   useTheme,
+  Throbber,
 } from '@embeddedchat/ui-elements';
 import { useLoginStore } from '../../store';
 import { useRCAuth } from '../../hooks/useRCAuth';
@@ -22,7 +23,7 @@ export default function LoginForm() {
   const setIsLoginModalOpen = useLoginStore(
     (state) => state.setIsLoginModalOpen
   );
-  const { handleLogin } = useRCAuth();
+  const { handleLogin, loading } = useRCAuth();
 
   const { theme } = useTheme();
 
@@ -132,11 +133,12 @@ export default function LoginForm() {
           <Button
             type="primary"
             onClick={handleSubmit}
+            disabled={loading}
             css={css`
               margin: 10px 0;
             `}
           >
-            Login
+            {loading ? <Throbber /> : 'Login'}
           </Button>
         </Box>
       </GenericModal>
--- a/packages/react/src/hooks/useRCAuth.js
+++ b/packages/react/src/hooks/useRCAuth.js
@@ -1,4 +1,4 @@
-import { useContext } from 'react';
+import { useContext, useState } from 'react';
 import { useToastBarDispatch } from '@embeddedchat/ui-elements';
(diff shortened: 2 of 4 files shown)


packages/react/src/views/LoginForm/LoginForm.js
```diff
@@ -43,7 +45,10 @@ export default function LoginForm() {
   const handleSubmit = () => {
     if (!userOrEmail) setUserOrEmail('');
     if (!password) setPassword('');
-    handleLogin(userOrEmail, password);
+    setLoading(true);
+    handleLogin(userOrEmail, password).finally(() => {
```
Spiral-Memory: Can we have it differently ? Instead of using finally here

Abbas-Askari: @Spiral-Memory 
We can put the `loading` in `useRCAuth` hook. Then the inside LoginForm we will have:

```javascript
const { handleLogin, loading } = useRCAuth();
...
const handleSubmit = () => {
    if (!userOrEmail) setUserOrEmail('');
    if (!password) setPassword('');
    handleLogin(userOrEmail, password); // No more setting loading to true and flase here.
  };
```
Would that be preferable?

I dont think you meant `try ... catch` instead of `finally`?

Abbas-Askari: > @Spiral-Memory We can put the `loading` in `useRCAuth` hook. Then the inside LoginForm we will have:
> 
> ```js
> const { handleLogin, loading } = useRCAuth();
> ...
> const handleSubmit = () => {
>     if (!userOrEmail) setUserOrEmail('');
>     if (!password) setPassword('');
>     handleLogin(userOrEmail, password); // No more setting loading to true and flase here.
>   };
> ```
> 
> Would that be preferable?
> 
> I dont think you meant `try ... catch` instead of `finally`?

Done. I also added loading indication to TOTP form.

Abbas-Askari:
Requested changes are completed @Spiral-Memory

PR #975, opened 2025-02-10
fix: Search Messages crashes on 429 from server #974

### LINKED ISSUE #974: bug: Search Messages crashes on 429 from server
### Description:

`TypeError: Cannot read properties of undefined (reading 'length')` appears when too many search requests are sent back to back in Search Messages Sidebar

### Steps to reproduce:

1. Goto https://rocketchat.github.io/EmbeddedChat/?path=/story/embeddedchat-simple--simple
2. Open `Search Messages` sidebar
3. Search repeatedly until the server returns `429`.
4. Observer the `TypeError: Cannot read properties of undefined (reading 'length')`

### Expected behavior:
No error should appear.

### Actual behavior:
Network tab on the right side is showing requests:

https://github.com/user-attachments/assets/15fb6766-ca64-4b2d-bd2b-8f5dee867afb

Abbas-Askari:
Fixes #974 by introducing a timeout (20s) on a 429 by the server and then trying again. In this period (timeouts and fetches) a `Throbber` (loading component) is rendered.

## Video/Screenshots
[Screencast from 2025-02-11 00-04-16.webm](https://github.com/user-attachments/assets/22b7d222-2f64-4d79-af63-0c951977a4c9)

## PR Test Details
**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-975 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.

### DIFF
--- a/packages/react/src/views/MessageAggregators/SearchMessages.js
+++ b/packages/react/src/views/MessageAggregators/SearchMessages.js
@@ -10,14 +10,26 @@ const SearchMessages = () => {
   const { RCInstance } = useContext(RCContext);
   const [text, setText] = useState('');
   const [messageList, setMessageList] = useState([]);
+  const [cooldown, setCooldown] = useState(false);
+  const [loading, setLoading] = useState(false);
 
   const handleInputChange = (e) => {
     setText(e.target.value);
   };
 
   const searchMessages = useCallback(async () => {
-    const { messages } = await RCInstance.getSearchMessages(text);
-    setMessageList(messages);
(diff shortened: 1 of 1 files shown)


Abbas-Askari:
There is one possible way of knowing the exact time to wait, But that is very messy and RC main dosen't use it as well. It will require the modification of `EmbeddedChatApi`. Instead of sending to `chat.search` we would send to `rocketchatSearch.search` which gives the exact milliseconds to wait. But its usage is completely different.

PR #1001, opened 2025-04-04
fix: React-Native project fails to load up during development

### LINKED ISSUE #1000: Version Mismatch: @embeddedchat/api Not Recognized in Monorepo Workspace
### Description:

There is a version mismatch between the @embeddedchat/api package in the monorepo and the published version on npm. The monorepo's package.json lists the version as 0.1.2, while the npm registry shows 0.1.3. This is causing issues when using yarn web in the react-native repo, as it fails to detect the local workspace package and instead tries to fetch it from npm.

### Steps to reproduce:

1. Run yarn web in the monorepo.
2. Observe the error: Error: @embeddedchat/api@workspace:^: Workspace not found (@embeddedchat/api@workspace:^).
3. Install the package from npm instead (0.1.3), and it works without issues.

### Expected behavior:

yarn web should correctly resolve @embeddedchat/api from the monorepo workspace instead of looking for the npm version and start the react-native project

### Actual behavior:

Yarn fails to find @embeddedchat/api@workspace:^ and throws an error.

Image

Image

**From the above snaps we can observe that the version used in the current api package is deprecated**

https://github.com/user-attachments/assets/0f07f7b9-15fb-48b6-8d47-2b02dcb4c4f3

SinghaAnirban005:
# React-Native project development fails

- [x]  Changed the version of the @embeddedchat/api package to use the latest one instead of the deprecated version
- [x] Ran yarn install at the root level 
- [x]  Hardcoded the api version inside the react-native repo

## Reason I have hardcoded instead of using workspace:^
1. While using workspace:^ works only on removing the yarn.lock file present inside the react-native repo, but doing this introduces dependency issues
2. The Metro bundler can't properly resolve the local dependency paths
3. Hence harcoding the version of api package inside the react-native repo is the best workaround to have for now

Fixes #1000 

## Video/Screenshots

https://github.com/user-attachments/assets/f2c64154-b2d2-43bc-b053-1e990411c82e

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-1001 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.


Spiral-Memory:
Hi @SinghaAnirban005 
Thanks for this PR

Why it is showing these many changes ?

Spiral-Memory:
Can you please ignore those yarn cache files

Spiral-Memory:
Hi @SinghaAnirban005 
Any idea, why yarn.lock is getting affected this much with just one version change

SinghaAnirban005:
Might be because of including a new library in the react-native package !!
I shall investigate further and then  let you know @Spiral-Memory

Spiral-Memory:
Sure, meanwhile I'm approving it

Spiral-Memory (approved):

SinghaAnirban005:
@Spiral-Memory 
The reason for such large number of additions accounts to the hardcoding of the version for @embeddedchat/api in the react-native project .  Due to which all the dependencies were installed separately, even if they exist already in the root or other package.

This can be tackled by using **workspace:^** instead of hardcoding version number . Due to hardcoding version number it was searching for the api package in the npm registry

But few considerations before using workspace:^ :
1. On running yarn I ran into an issue where Yarn couldn’t resolve the workspace version and threw an error like **could not find workspace:^.**
After digging in, I found that this was caused by the presence of an existing yarn.lock file inside packages/react-native. This lockfile caused Yarn to treat it as an external package instead of linking the local workspace version.

2. To resolve this i removed the old yarn.lock file and ran yarn install . After that, things worked as expected.

So the solution given in the PR can  be thought of as a temporary workaround . I didn't want to delete the existing yarn.lock file so i had to come up with the given solution

Spiral-Memory:
Then do one think, delete that, run yarn, to get a new lock file and then we can have that 

Will that work ?

SinghaAnirban005:
deleting the yarn.lock file inside the native repo and then running a yarn install wont generate a new lockfile since it is treated as a yarn workspace

i believe there should be only a central lockfile for this monorepo at the root level

abirc8010:
> While using workspace:^ works only on removing the yarn.lock file present inside the react-native repo, but doing this introduces dependency issues

I’ve tested this as well. Using workspace:^ in embeddedchat/api leads to dependency issues, unless the expo package is installed at the root level. Additionally, to make it compatible, some more React Native specific packages also need to be added to the monorepo’s dependencies.

SinghaAnirban005:
Absolutely !
I felt that this issue can be solved more comprehensively during the Gsoc period since it involves upgrading and importing of few packages (aligns with the deliverables in the Emb chat 2025) hence my solution was just to provide a temporary workaround for few contributors who were complaining about this issue.
I have confirmed with Zishan that this particular issue can later be handled more gracefully by whoever is selected for this project .
Thank You

Spiral-Memory:
@SinghaAnirban005 @abirc8010 @devanshkansagra 
If anyone of you is available, look into it and see if we can atleast make that mobile app in a working condition atleast with minimal changes. Thanks

PR #1009, opened 2025-06-09
Refactor: Migrating from RocketChat SDK to DDPSDK of Rocket.Chat

devanshkansagra:
# Proposed changes

This Pull Request updates the EmbeddedChat API and EmbeddedChat Auth to use the new DDPSDK of RocketChat instead of the Rocket.Chat SDK.

Fixes: N/A

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-1009 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.

PR #1033, opened 2025-12-05
fixing messages not being rendered inside threads

### LINKED ISSUE #1032: FIX : Threads messages fetching endpoint populated with unnecessary parameters.
### Description:

When a user decides to convert a message to a thread for more discussions on the topic, messages are being saved which are sent in the thread but when tried to fetch the thread messages, backend throws an error which says that there were extra unnecessary parameters included in the API endpoint URL.

### Steps to reproduce:

1. Go to Embedded chat.
2. Type a message and click on the threads option.
3. Enter a message in the thread.
4. Now try to go back and come back to this thread.
5. When you check the network section, you can see the backend throws the error as a result of which the messages sent in the threads are not shown to the end user.

### Expected behavior:

The threads messages should render correctly.

### Actual behavior:

Currently we are hitting the right endpoint but with an extra query parameter which is causing the issue as show in the below snip shot.

Image

https://github.com/user-attachments/assets/90d7b384-9519-48a8-9496-24198c30ea29

Endpoint being hit:

`http://localhost:3000/api/v1/chat.getThreadMessages?roomId=GENERAL&tmid=ec_5ee90c109e3f4b03b`
Based on the Rocket.chat code, there is no necessity of the roomId in the query parameter.

Pratheek555:
# Brief Title

Fixing the unnecessary query parameters for the getThreadMessages endpoint which is used to fetch the messages of  respective threads.

Fixes #1032 

## Video/Screenshots
image

Note: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-1033 after approval. Contributors are requested to replace <pr_number> with the actual PR number.

### DIFF
--- a/packages/api/src/EmbeddedChatApi.ts
+++ b/packages/api/src/EmbeddedChatApi.ts
@@ -632,7 +632,7 @@ export default class EmbeddedChatApi {
     try {
       const { userId, authToken } = (await this.auth.getCurrentUser()) || {};
       const messages = await fetch(
-        `${this.host}/api/v1/chat.getThreadMessages?roomId=$tmid=${tmid}`,
+        `${this.host}/api/v1/chat.getThreadMessages?tmid=${tmid}`,
         {
           headers: {
             "Content-Type": "application/json",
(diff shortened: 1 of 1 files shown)


Spiral-Memory:
Thanks for consistently fixing bugs as Rocket.Chat new API evolves. Everything was working before, but it seems the recent changes in the API might be causing these issues. It might be a good idea to create a main issue to update it in line with Rocket.Chat’s APIs and include Rocket.Chat DDP, as shown in this PR: https://github.com/RocketChat/EmbeddedChat/pull/1009

Spiral-Memory (approved):

PR #1048, opened 2026-01-01
Fix/embedded chat channelname not switching

### LINKED ISSUE #452: Change current channel with Channel Name
As an EmbeddedChat developer

**I need to**:
Implement the feature to change the current channel with `Channel Name`.
For now, it is only possible to do so with `Channel ID` but it is difficult for many to find it, so allowing Channel Name to also change the current channel would be a great feature.

**Acceptance Criteria**
- [ ] Allow Channel Name to also change the current active channel tab

deepak0x:
# Enable Channel Switching by Channel Name

Closes #452

## Overview

This PR adds the ability to switch channels using channel names instead of requiring room IDs. Previously, users had to find and provide the room ID to switch channels, which was difficult. Now, users can simply provide the channel name (e.g., "public", "embedded", "general") and the component will automatically resolve it to the correct room ID.

## Problem

Before this change, the `EmbeddedChat` component only supported switching channels using `roomId`. Channel IDs are not user-friendly and difficult to find, making it hard for developers to integrate the component with specific channels.

## Solution

The component now supports channel switching using the `channelName` prop. When `channelName` is provided:

1. The component waits for user authentication
2. Once authenticated, it calls the Rocket.Chat API to resolve the channel name to a room ID
3. The resolved room ID is used to connect to the correct channel
4. Messages and channel information are properly loaded for the resolved channel

### Core Functionality

- Added `resolvedRoomId` state to manage dynamically resolved room IDs
- Added `explicitRoomId` state to track when a room ID is explicitly provided
- Implemented channel name to room ID resolution using `/api/v1/rooms.info?roomName=...` API endpoint
- Updated re-instantiation logic to properly clear messages when switching channels
- Added proper error handling for authentication and API failures

### UI Updates

- Updated `ChatHeader` to prioritize `channelName` prop for display
- Updated `ChatInput` placeholder to use `channelName` when available
- Added error handling for unauthorized channel info requests

### API Changes

- Updated `EmbeddedChatApi.channelInfo()` to check for authentication before making API calls
- Returns proper error response when authentication is not available

### Using Channel Name (New)

```jsx
<EmbeddedChat
  host="http://localhost:3000"
  channelName="public"
  auth={{
    flow: 'PASSWORD'
  }}
/>
```

### Using Room ID (Still Supported)

```jsx
<EmbeddedChat
  host="http://localhost:3000"
  roomId="6956b4ced42fcf35f5783b1d"
  channelName="public"
  auth={{
    flow: 'PASSWORD'
  }}
/>
```

## How It Works

1. If `roomId` is explicitly provided, it is used directly
2. If only `channelName` is provided:
   - Component waits for user authentication
   - Once authenticated, makes API call to resolve channel name to room ID
   - Uses resolved room ID to connect to the channel
3. If neither is provided, defaults to "GENERAL" channel

## Before / After

<table>
<tr>
<th>Before</th>
</tr>
<tr>
<td align="center">
before
</td>
</tr>

<tr>
<th>After (with channel name)</th>
</tr>
<tr>
<td align="center">
after channel name
</td>
</tr>

<tr>
<th>After (with room ID)</th>
</tr>
<tr>
<td align="center">
after room id
</td>
</tr>
</table>

## Technical Details

- The resolution happens asynchronously after authentication
- The component properly handles cases where authentication is not yet complete
- Messages and channel state are cleared when switching channels to prevent showing wrong channel data
- The component re-instantiates the connection when the resolved room ID changes

## Testing

- Tested with channel names: "public", "embedded", "general"
- Verified that channel switching works correctly with only channel name provided
- Verified that explicit room ID still works as before
- Verified that messages are properly cleared when switching channels
- Verified that authentication errors are handled gracefully

### DIFF
--- a/packages/react/src/hooks/useRoomId.js
+++ b/packages/react/src/hooks/useRoomId.js
@@ -30,7 +30,7 @@ export const useRoomId = (
       return { roomId, error: null };
     }
     return channelName
-      ? { roomId: null, error: null }
+      ? { roomId: 'GENERAL', error: null }
       : { roomId: 'GENERAL', error: null };
   });
 
@@ -43,6 +43,7 @@ export const useRoomId = (
 
       if (channelName) {
         if (!isUserAuthenticated) {
+          setResolvedRoomId({ roomId: 'GENERAL', error: null });
           return;
         }
 
--- a/packages/react/src/hooks/useRoomId.js
+++ b/packages/react/src/hooks/useRoomId.js
@@ -1,0 +1,92 @@
+import { useState, useEffect } from 'react';
+import { EmbeddedChatApi } from '@embeddedchat/api';
+
+/**
+ * Custom hook to resolve roomId from either explicit roomId or channelName.
+ * Returns an object with:
+ * - roomId: The resolved room ID, or null if resolution is pending/failed
+ * - error: Error message if resolution failed, or null if successful/pending
+ *
+ * @param {string|null|undefined} roomId - Explicit room ID
+ * @param {string|null|undefined} channelName - Channel name to resolve
+ * @param {string} host - Rocket.Chat host URL
+ * @param {Function} getToken - Token getter function
+ * @param {Function} deleteToken - Token deleter function
+ * @param {Function} saveToken - Token saver function
+ * @param {boolean} isUserAuthenticated - Whether user is authenticated
+ * @returns {{roomId: string|null, error: string|null}} - Resolved room ID and error state
+ */
+export const useRoomId = (
+  roomId,
+  channelName,
+  host,
+  getToken,
+  deleteToken,
+  saveToken,
+  isUserAuthenticated
+) => {
+  const [resolvedRoomId, setResolvedRoomId] = useState(() => {
+    if (roomId) {
+      return { roomId, error: null };
+    }
+    return channelName
+      ? { roomId: null, error: null }
+      : { roomId: 'GENERAL', error: null };
+  });
+
+  useEffect(() => {
+    const resolveRoomId = async () => {
+      if (roomId) {
+        setResolvedRoomId({ roomId, error: null });
+        return;
+      }
+
+      if (channelName) {
+        if (!isUserAuthenticated) {
+          return;
+        }
+
+        try {
+          const tempRCInstance = new EmbeddedChatApi(host, 'GENERAL', {
+            getToken,
+            deleteToken,
+            saveToken,
+          });
+          const roomIdFromName = await tempRCInstance.getRoomIdByName(
+            channelName
+          );
+          await tempRCInstance.close().catch(console.error);
+          if (roomIdFromName) {
+            setResolvedRoomId({ roomId: roomIdFromName, error: null });
+          } else {
+            setResolvedRoomId({
+              roomId: null,
+              error: `Channel "${channelName}" not found or you don't have access to it.`,
+            });
+          }
+        } catch (error) {
+          setResolvedRoomId({
+            roomId: null,
+            error: `Failed to resolve channel "${channelName}": ${
+              error.message || 'Unknown error'
+            }`,
+          });
+        }
+      } else {
+        setResolvedRoomId({ roomId: 'GENERAL', error: null });
+      }
+    };
+
+    resolveRoomId();
+  }, [
+    roomId,
+    channelName,
+    host,
+    getToken,
+    deleteToken,
+    saveToken,
+    isUserAuthenticated,
+  ]);
+
+  return resolvedRoomId;
+};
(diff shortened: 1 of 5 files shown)


packages/react/src/views/EmbeddedChat.js
```diff
@@ -27,9 +32,23 @@ import { overrideECProps } from '../lib/overrideECProps';
 
 const EmbeddedChat = (props) => {
   const [config, setConfig] = useState(() => props);
+  // Track if roomId was explicitly provided (not just the default)
```
Spiral-Memory: Please remove all the comments

deepak0x:
@Spiral-Memory I removed all the comments...

deepak0x:
@Spiral-Memory I fixed that yarn lint issue 
was coming due to empty block

deepak0x:
@Spiral-Memory fixed the EmbeddedChatApi.ts to pass format check

packages/react/src/views/EmbeddedChat.js
```diff
@@ -90,34 +105,144 @@ const EmbeddedChat = (props) => {
...
+              return;
+            }
+            throw new Error(`HTTP error! status: ${response.status}`);
+          }
+
+          const data = await response.json();
+          if (data?.success && data?.room?._id) {
+            setResolvedRoomId(data.room._id);
+          } else {
+            setResolvedRoomId('GENERAL');
```
Spiral-Memory: Do we really need to have this fallback ?

packages/react/src/views/EmbeddedChat.js
```diff
@@ -90,34 +105,144 @@ const EmbeddedChat = (props) => {
...
+        }
+      } else {
+        setResolvedRoomId('GENERAL');
+      }
+    };
+
+    resolveRoomId();
+  }, [channelName, explicitRoomId, host, RCInstance, isUserAuthenticated]);
+
+  useEffect(() => {
```
Spiral-Memory, line 121: Can you please explain the use for this

packages/react/src/views/EmbeddedChat.js
deepak0x: I set this If a user provides a channelName that doesn't exist or they lack access, falling back to 'GENERAL' hides the failure.

packages/react/src/views/EmbeddedChat.js
deepak0x: This useEffect resolves a channelName to a roomId (e.g., "GENERAL", "6956b4ced42fcf35") when only the channel name is provided.

Spiral-Memory:
Hi @deepak0x  
This approach seems to involve too much manual work, so I wouldn’t recommend it.  

`async getRoomIdByName(channelName: string): Promise<string> {` could be part of the API instead.  

Rather than manually handling hasMounted, we could use a key-based approach. A custom hook, like useRoomId or something similar, could also be created.  

Please look into finding a more elegant and efficient solution.

deepak0x:
Sure @Spiral-Memory

deepak0x:
Hey @Spiral-Memory 
Have addressed issues reported by you..  pls check...

deepak0x:
hey @Spiral-Memory 
If u get time you can check this pr 
have fixed the logic as per previous description pls check and let me know if any issue
Thanks ;)

packages/react/src/hooks/useRoomId.js
```diff
@@ -0,0 +1,93 @@
+import { useState, useEffect } from 'react';
+import { EmbeddedChatApi } from '@embeddedchat/api';
+
+/**
+ * Custom hook to resolve roomId from either explicit roomId or channelName.
```
Spiral-Memory, line 5: Please remove the comments

packages/react/src/hooks/useRoomId.js
```diff
@@ -0,0 +1,93 @@
...
+ * Returns an object with:
+ * - roomId: The resolved room ID, or null if resolution is pending/failed
+ * - error: Error message if resolution failed, or null if successful/pending
+ *
+ * @param {string|null|undefined} roomId - Explicit room ID
```
Spiral-Memory, line 10: Please remove the definitions

packages/react/src/views/EmbeddedChat.js
```diff
@@ -89,37 +87,69 @@ const EmbeddedChat = (props) => {
...
+    host,
+    getToken,
+    deleteToken,
+    saveToken,
+    isUserAuthenticated
+  );
 
-    return newRCInstance;
-  }, [host, roomId, getToken, deleteToken, saveToken]);
+  const RCInstance = useMemo(() => {
```
Spiral-Memory, line 101: I'm still concerned that it involves too much manual calculation and potential risks if merged. Could you remove anything redundant and create a video demonstrating a full end-to-end test of all the features?

deepak0x: Sure

Spiral-Memory:
Closing this PR as it looks stale.

PR #1052, opened 2026-01-04
fix: replace import assertion with readFileSync in rollup configs

### LINKED ISSUE #1050: Build fails with "SyntaxError: Unexpected identifier 'assert'" in `@embeddedchat/auth` package
The build process fails when building the `@embeddedchat/auth` package due to Rollup not parsing the import assertion syntax in `rollup.config.js`.

## Error Message

Image

```
✖  @embeddedchat/auth:build
   [!] SyntaxError: Unexpected identifier 'assert'
       at compileSourceTextModule (node:internal/modules/esm/utils:344:16)
       at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:105:18)
       at ModuleLoader.#translate (node:internal/modules/esm/loader:534:12)
       at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:581:27)
       at ModuleJob._link (node:internal/modules/esm/module_job:116:19)
```

## Steps to Reproduce

1. Clone the repository
2. Install dependencies: `yarn`
3. Run build: `yarn build`
4. Build fails on `@embeddedchat/auth` package

## Root Cause

The issue is in `packages/auth/rollup.config.js` line 4:

```javascript
import packageJson from './package.json' assert { type: 'json' };
```

Rollup's parser doesn't support the ESM import assertion syntax (`assert { type: 'json' }`) when processing the config file, even though Node.js supports it.

## Affected Files

- `packages/auth/rollup.config.js` (line 4)
- `packages/api/rollup.config.js` (line 3) - has the same issue

## Environment

- **Node.js Version**: v22.16.0
- **Package Manager**: Yarn
- **Build Tool**: Rollup
- **OS**: macOS

## Expected Behavior

The build should complete successfully for all packages.

deepak0x:
- Replace 'import packageJson from ./package.json assert { type: 'json' }' with readFileSync
- Fixes build error: SyntaxError: Unexpected identifier 'assert'
- Affects packages/auth and packages/api rollup.config.js files
- Uses Node.js fs.readFileSync which is compatible with Rollup parser

Closes #1050

Mac Recording
https://github.com/user-attachments/assets/257f6cb5-2632-4da9-9157-99849ca26934

### DIFF
--- a/packages/api/rollup.config.js
+++ b/packages/api/rollup.config.js
@@ -1,7 +1,13 @@
 import dts from 'rollup-plugin-dts'
 import esbuild from 'rollup-plugin-esbuild'
 import path from 'path';
-import packageJson from './package.json' with { type: 'json' };
+import { createRequire } from 'module';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const require = createRequire(import.meta.url);
+const packageJson = require(path.resolve(__dirname, './package.json'));
 
 const name = packageJson.main.replace(/\.js$/, '');
 
--- a/packages/api/rollup.config.js
+++ b/packages/api/rollup.config.js
@@ -1,9 +1,7 @@
 import dts from 'rollup-plugin-dts'
 import esbuild from 'rollup-plugin-esbuild'
 import path from 'path';
-import { readFileSync } from 'fs';
-
-const packageJson = JSON.parse(readFileSync('./package.json', 'utf-8'));
+import packageJson from './package.json' with { type: 'json' };
 
 const name = packageJson.main.replace(/\.js$/, '');
 
--- a/packages/api/rollup.config.js
+++ b/packages/api/rollup.config.js
@@ -1,7 +1,9 @@
 import dts from 'rollup-plugin-dts'
 import esbuild from 'rollup-plugin-esbuild'
-import packageJson from './package.json' assert { type: 'json' };
 import path from 'path';
+import { readFileSync } from 'fs';
+
+const packageJson = JSON.parse(readFileSync('./package.json', 'utf-8'));
 
 const name = packageJson.main.replace(/\.js$/, '');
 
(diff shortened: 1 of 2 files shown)


deepak0x:
@Spiral-Memory can you please check this pr...

Spiral-Memory (approved):
LGTM, thanks for the fix @pratheek56. If possible, could you attach a video showing the build succeeding on Linux, as well as on Windows and Mac if available.

deepak0x:
sure @Spiral-Memory

Spiral-Memory:
Instead of using reading, which could be a blocking operation, we could use with { type: 'json' } instead of assert. I think that would work just as well and be non-blocking. Please look into that option.

deepak0x:
actually i check and got that readfilesync is good option..
{ type: 'json' } will work too...

Spiral-Memory:
Please change to with { type: 'json' }

Spiral-Memory (approved):
Thanks @deepak0x 
Hope build is succeeding ?

Spiral-Memory:
It seems this solution isn’t backward compatible with Node 16. Consider trying and explore similar solution:

```
const require = createRequire(import.meta.url);
const packageJson = requirepackage.json');
```

This approach provides module caching and ensures backward compatibility.

deepak0x:
Hi @Spiral-Memory 
The preinstall script requires Node.js v16.19.0.. causing failure during preinstall
should i change this requirement to 16 or greater?

Spiral-Memory:
If we switch to Node 16 or higher, it might cause issues for those still using Node 16. That would mean updating the documentation to require Node > 16, but we haven’t fully tested on versions above 16 yet. I’m looking a solution that’s backward compatible, supports module caching (unlike your initial approach), and works with higher Node versions as well.

deepak0x:
@Spiral-Memory i addresses your issue 
pls check

Spiral-Memory (approved):
LGTM

deepak0x:
added mac recording...

Spiral-Memory (approved):

Spiral-Memory:
Thanks for the fix @deepak0x

PR #1077, opened 2026-01-17
fix: resolve memory leak in MediaRecorder event listeners

### LINKED ISSUE #1073: Bug: Event Listener Memory Leak in useMediaRecorder Hook
The `useMediaRecorder` and `useNewMediaRecorder` hooks attach event listeners to `MediaRecorder` instances using `addEventListener`, but never remove them. When recording is started multiple times, new `MediaRecorder` instances are created while old listeners remain attached to abandoned instances.

This leads to memory leaks and increasing memory usage over time.

---
## Observed Behavior

Both hooks register event listeners on `MediaRecorder` objects using `addEventListener`:

* `dataavailable`
* `stop`

These listeners are added every time recording starts. However, they are **never removed** when recording stops or when a new recorder instance is created.

As a result, listeners accumulate across multiple recording sessions.

---

## Why This Is Incorrect

Event listeners must be explicitly removed when an object is no longer needed. In this case:

* `MediaRecorder` instances are recreated on each recording session
* Old recorder instances are abandoned
* Event listeners remain attached to those instances

This prevents garbage collection and causes memory leaks.

---

## Steps to Reproduce

1. Use the audio or video recording feature in EmbeddedChat.
2. Start recording.
3. Stop recording.
4. Start recording again (a new `MediaRecorder` instance is created).
5. Repeat steps 2–4 multiple times.
6. Observe browser memory usage using developer tools.

Memory usage increases steadily due to unreleased `MediaRecorder` instances.

deepak0x:
This PR fixes a memory leak in the useMediaRecorder and useNewMediaRecorder hooks where event listeners attached to MediaRecorder instances were never removed, causing unused recorder instances to remain in memory after each recording session.

Closes #1073

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-1077 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.

### DIFF
--- a/packages/react/src/hooks/useMediaRecorder.js
+++ b/packages/react/src/hooks/useMediaRecorder.js
@@ -28,19 +28,29 @@ export function useMediaRecorder({ constraints, onStop, videoRef }) {
     const stream = await getStream(constraints, true);
     chunks.current = [];
     const _recorder = new MediaRecorder(stream);
+
+    const handleDataAvailable = (event) => {
+      chunks.current.push(event.data);
+    };
+
+    const handleStop = () => {
+      onStop && onStop(chunks.current);
+      _recorder.removeEventListener('dataavailable', handleDataAvailable);
+      _recorder.removeEventListener('stop', handleStop);
+    };
+
+    _recorder.addEventListener('dataavailable', handleDataAvailable);
+    _recorder.addEventListener('stop', handleStop);
+
     _recorder.start();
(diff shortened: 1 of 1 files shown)


Spiral-Memory (approved):
LGTM

Spiral-Memory (approved):

PR #1089, opened 2026-01-19
feat: add clickable timestamp to quoted messages

### LINKED ISSUE #1088: Quoted messages do not provide a way to navigate to the original message
### Description:
When a message is quoted, the UI displays only the author name and quoted text.
There is no way for users to navigate back to the original message from the quote.

Why this is a problem

- In long conversations, users often lose context
- Users expect quotes to behave like references
- Other chat apps allow quick navigation to the original msg, and Rocket.Chat core already provides similar navigation behavior.

### Steps to reproduce:

1. Quote a message.
2. Observer there is no reference.
3. There is no clickable element to navigate to the original message

### Actual behavior:

**There is no way to jump the original message.**

https://github.com/user-attachments/assets/f2375f91-df58-48e3-81a4-c2cfac1c04e6

### Expected behavior:

**Users should be able to click something (timestamp) in the quote to jump to the original message if it exists.**

https://github.com/user-attachments/assets/17b36867-3d36-4a90-b23b-0a81abc54495

Khizarshah01:
# Add clickable timestamp to quoted messages

- [x] Display a timestamp next to the quoted message author
- [x] Make the timestamp clickable to navigate to the original message

Fixes #1088 

## Video/Screenshots

https://github.com/user-attachments/assets/a02ff8a8-1fb8-4d61-85d9-b396f4da1326

## Technical notes

- This PR adds navigation to the original quoted message by locating the message element in the DOM using its message ID and scrolling to it when available.

- **Navigation works only when the original message is already loaded and rendered in the DOM**

- I explored whether there is an existing way to navigate to or load messages by ID in EmbeddedChat, but couldn’t find a suitable mechanism. If there is a recommended approach, I’d be happy to update the implementation

## PR Test Details

### DIFF
--- a/packages/react/src/views/AttachmentHandler/TextAttachment.js
+++ b/packages/react/src/views/AttachmentHandler/TextAttachment.js
@@ -1,4 +1,5 @@
 import React, { useState, useContext } from 'react';
+import { format } from 'date-fns';
 import { css } from '@emotion/react';
 import PropTypes from 'prop-types';
 import { Box, Avatar, useTheme, Icon } from '@embeddedchat/ui-elements';
@@ -11,12 +12,11 @@ const FileAttachment = ({
   attachment,
   host,
   type,
-  author,
   variantStyles = {},
   msg,
 }) => {
   const { RCInstance } = useContext(RCContext);
-  const { theme } = useTheme();
+  const { theme, mode } = useTheme();
   const [isExpanded, setIsExpanded] = useState(true);
 
   const getUserAvatarUrl = (icon) => {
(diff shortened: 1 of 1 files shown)


Spiral-Memory:
It's a great improvement, but currently it will not work properly if the message is not in the DOM. We need to implement proper virtualization of the message chat to prevent jumping behaviors. @deepak0x is working on it, so please coordinate with him to achieve this.

Meanwhile, I'll review this PR. It will at least work when the message is loaded in the DOM.

Spiral-Memory (approved):
LGTM

Khizarshah01:
Since #1111 introduces react-virtuoso, I can adapt the quote navigation to use scrollToIndex once **merged**

PR #1090, opened 2026-01-20
fix: infinite render loop in media recorders by adding missing depend…

### LINKED ISSUE #1083: Bug: Performance Degradation: Infinite Render Loop in Media Recorders
A performance issue exists in the audio and video message recorder components where an infinite render loop is triggered due to incorrect usage of `useCallback` and `useEffect`.

The `handleMount` callback is recreated on every render, which causes a dependent `useEffect` to re-run continuously. This results in repeated async operations, excessive CPU usage, and severe performance degradation.

---

## Observed Behavior

* `handleMount` is defined using `useCallback` without a dependency array.
* Because of this, a new function reference is created on every render.
* `handleMount` is included in the dependency array of a `useEffect`.
* The effect runs on every render and invokes `handleMount`.
* Internal state updates inside `handleMount` trigger another render.

This creates an infinite loop of renders and async calls.

Image
---

## Why This Is Incorrect

* `useCallback` without a dependency array provides no memoization.
* A new function reference on each render defeats the purpose of `useCallback`.
* When such a function is used as a `useEffect` dependency, it causes the effect to re-run continuously.
* 
---

## Steps to Reproduce

1. Open the Chat component.
2. Open browser DevTools and monitor CPU usage.
3. Add a log inside `handleMount`.
4. Observe that `handleMount` is called repeatedly.

Result:

* `handleMount` executes thousands of times per second.
* CPU usage remains constantly high.

---

## Expected Behavior

* `handleMount` should be memoized properly.
* If it only needs to run once, it should use an empty dependency array.

deepak0x:
This pull request fixes a severe performance issue where the `AudioMessageRecorder` and `VideoMessageRecorder` components were entering an infinite render loop, leading to high CPU usage and unstable UI behavior.

---

## Root Cause

The `handleMount` function in both components was defined using `useCallback` **without a dependency array**.

In React:

* A `useCallback` without dependencies returns a new function reference on every render.
* Both components used `handleMount` as a dependency of a `useEffect` that invokes it.

This caused the following loop:

1. Component renders
2. New `handleMount` function reference is created
3. `useEffect` detects a dependency change and runs
4. `handleMount` executes
5. Internal state updates and permission/device queries trigger a re-render
6. The cycle repeats indefinitely

This issue was especially noticeable when typing in the `ChatInput` or during UI state updates.

---

Closes #1083

## PR Test Details

### DIFF
--- a/packages/react/src/views/ChatInput/AudioMessageRecorder.js
+++ b/packages/react/src/views/ChatInput/AudioMessageRecorder.js
@@ -119,7 +119,7 @@ const AudioMessageRecorder = (props) => {
     } catch (error) {
       console.warn(error);
     }
-  });
+  }, []);
 
   useEffect(() => {
     handleMount();
--- a/packages/react/src/views/ChatInput/VideoMessageRecoder.js
+++ b/packages/react/src/views/ChatInput/VideoMessageRecoder.js
@@ -86,7 +86,7 @@ const VideoMessageRecorder = (props) => {
     } catch (error) {
       console.warn(error);
     }
-  });
+  }, []);
 
   useEffect(() => {
     handleMount();


Spiral-Memory (approved):
LGTM. Thanks

deepak0x:
Thanks to you too

PR #1096, opened 2026-01-23
feat: implement offline message persistence and refined UI

### LINKED ISSUE #1091: Feature Request: Offline Message Handling and Resend/Delete UI
When a user sends a message while offline or when a network request fails, the application does not provide clear feedback or recovery options. Messages remain in a pending state indefinitely, cannot be retried or deleted through the UI, and are lost entirely on page reload.

This results in poor reliability, confusing user experience, and potential data loss.

---

## Current Behavior

1. User sends a message while offline or during a network failure.
2. The message appears in the chat list with a pending or faded style.
3. The network request fails or times out.
4. The message remains in a pending state indefinitely.
5. No error indication is shown.
6. No retry or delete action is available.
7. If the page is reloaded, the pending message disappears completely.

Screenshots

Image

Proposed UI

Image

Image

---

## Problems Identified

* No distinction between pending and failed messages
* No visual error feedback to the user
* No ability to retry or delete failed messages
* Failed and pending messages are stored only in memory
* Messages are lost on page reload

---

## Desired Behavior

When a message fails to send due to offline status or a network error:

* Message status should transition from `pending` to `error`
* An error icon (for example, an exclamation mark) should be displayed next to the message
* Clicking the error icon should present the following actions:

  * Resend: Retry sending the message
  * Delete: Remove the message from the local list

deepak0x:
This PR implements **robust Offline Message Persistence** along with a **refined UI for failed message delivery**. It ensures that messages sent while offline are not lost, allows users to recover them after a page reload, and provides a clear, intuitive interface for resending or deleting failed attempts.

Closes  #1091

---

Screenshots/Video

https://github.com/user-attachments/assets/aaecc0da-0647-4897-95ec-34813cf606be

### 1. Offline Persistence (localStorage)

* Implemented `localStorage`-based caching for messages that fail to send.
* Messages are partitioned by **Room ID (`rid`)**, ensuring they only reappear in their respective channels.
* Persisted messages are automatically re-injected into the message store during:

  * Initialization
  * Room switching
* Added session cleanup logic to **wipe offline caches on user logout**.

---

### 2. Refined Message Sorting

* Ensures global ordering consistency between:

  * Server-fetched messages
  * Locally recovered offline messages
* **Resend behavior:** When a message is resent, its timestamp is updated to the current time, naturally moving it to the bottom of the chat.

---

### 3. UX & UI Improvements

* **Offline Indicator:** Added a centered red `RiWifiOffLine` icon for messages that fail to send.
* **Custom Error Menu:** Clicking the error icon opens an upward-positioned menu with:

  * Resend
  * Delete


Spiral-Memory:
This looks good overall, but I’d prefer the Wi-Fi icon to be more understated - small and subtle, to give it a cleaner, more polished look.

packages/react/src/store/messageStore.js
```diff
@@ -2,8 +2,46 @@ import { create } from 'zustand';
 import cloneArray from '../lib/cloneArray';
 import { upsertMessage } from '../lib/messageListHelpers';
 
+const PERSISTENCE_KEY = 'ec_offline_messages';
```
Spiral-Memory: This can we store somewhere globally instead of messageStore

yarn.lock
```diff
@@ -26877,6 +26878,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"react-icons@npm:^5.5.0":
```
Spiral-Memory: Use only the available icons, and generate them through the built-in icon tool from Rocket.Chat’s core system.

deepak0x:
sure 
will fix and commit

deepak0x:
Hi @Spiral-Memory, addressed all three review comments:

**1. PERSISTENCE_KEY moved globally**
Created `packages/react/src/lib/constants.js` — exports `EC_OFFLINE_MESSAGES_KEY`. `messageStore.js` now imports it from there.

**2 & 3. Replaced `react-icons` with built-in icon + made it more understated**
Removed `RiWifiOffLine` from `react-icons/ri` entirely. Now using `<Icon name="error-circle" />` from `@embeddedchat/ui-elements` — smaller (`0.875em`), subtle opacity (`0.7`), uses `theme.colors.danger`.

Before vs After:

IMG_1822

Please let me know if any further changes are needed!

deepak0x:
Also removed `react-icons` package entirely from `packages/react/package.json` — it was the only usage and is now replaced by the built-in icon system. yarn.lock updated accordingly.

PR #1100, opened 2026-01-24
fix: resolve duplicated links when quoting multiple messages in ChatI…

vivekyadav-3:
While exploring the 
ChatInput
 component, I noticed that quoting multiple messages led to redundant content accumulation in the message body. This was caused by a string mutation inside the quoteMessage.map loop.

I've refactored the logic to use a clean functional approach (map + join), which ensures each quote correctly generates exactly one link and improves performance

### DIFF
--- a/packages/react/src/views/ChatInput/ChatInput.js
+++ b/packages/react/src/views/ChatInput/ChatInput.js
@@ -298,17 +298,17 @@ const ChatInput = ({ scrollToBottom, clearUnreadDividerRef }) => {
       //   }
       // }
 
-      const quoteArray = await Promise.all(
+      const quoteLinks = await Promise.all(
         quoteMessage.map(async (quote) => {
           const { msg, attachments, _id } = quote;
           if (msg || attachments) {
             const msgLink = await getMessageLink(_id);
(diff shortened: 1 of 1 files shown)


Spiral-Memory:
Could you share a video demonstrating the changes, showing both the state before and after the fixes were applied?

vivekyadav-3:
Hi @Spiral-Memory!
Thank you for your patience. Here are the screen recordings demonstrating the changes before and after my fix.
https://github.com/user-attachments/assets/3ff5845a-dd43-4064-b593-9e25c88253d9
In the first video, you can see that when quoting multiple messages, the old code in 
ChatInput.js ended up duplicating the URL links every single time it iterated over the quotes array. When creating the final message, it resulted in a messy output of stacked duplicate links.

https://github.com/user-attachments/assets/944ad102-23a5-40e8-8080-81bc406eaf43

With my PR applied, instead of joining everything in an outer variable inside the .map() loop, the code correctly returns each link individually and calls .join('') at the very end just once. Now, when I quote those exact same two messages and hit send, the output string is perfectly clean with no duplications.

PR #1102, opened 2026-01-24
fix(react): add missing unique key props to ChatInputFormattingToolba…

### LINKED ISSUE #1098: [Bug]: Missing unique "key" prop in ChatInputFormattingToolbar components
A React console warning occurs when the message formatting toolbar is rendered. React requires every item in a list (such as the icons in the toolbar) to have a unique `key` prop so it can efficiently track updates. Currently, several components in the toolbar are missing these keys, which triggers the following warning:

```
Warning: Each child in a list should have a unique "key" prop. Check the render method of ChatInputFormattingToolbar.
```
Screenshots/Video

Image

---

## Steps to Reproduce

1. Open the EmbeddedChat application or Storybook.
2. Navigate to any view where the Chat Input Toolbar is visible.
3. Open the browser Developer Console.
4. Observe the warning during the initial render or when interacting with the toolbar.

deepak0x:
This PR fixes a common React warning:

```
Warning: Each child in a list should have a unique "key" prop
```

The warning occurred in the `ChatInputFormattingToolbar` component and was caused by several elements being rendered as part of dynamic lists without providing a unique `key` prop. This made it harder for React to correctly track and optimize updates to these elements.

---

Closes #1098

## Changes

Updated `packages/react/src/views/ChatInput/ChatInputFormattingToolbar.js`:

* Added `key="audio"` to the `AudioMessageRecorder` component.
* Added `key="video"` to the `VideoMessageRecorder` component.
* Replaced empty fragments (`<>`) with `<React.Fragment key={item.name}>` for mapped formatting buttons (bold, italic, strike, etc.).

---

### DIFF
--- a/packages/react/src/views/ChatInput/ChatInputFormattingToolbar.js
+++ b/packages/react/src/views/ChatInput/ChatInputFormattingToolbar.js
@@ -116,6 +116,7 @@ const ChatInputFormattingToolbar = ({
 
     audio: (
       <AudioMessageRecorder
+        key="audio"
         displayName={
           isPopoverOpen && popOverItems.includes('audio') ? 'audio' : null
         }
@@ -125,6 +126,7 @@ const ChatInputFormattingToolbar = ({
     ),
     video: (
       <VideoMessageRecorder
+        key="video"
         displayName={
           isPopoverOpen && popOverItems.includes('video') ? 'video' : null
         }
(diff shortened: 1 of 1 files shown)


sudorishabh:
Bro, I saw the same issue earlier and searched to see if there was already an open PR. I came across your implementation. Great work. However, there’s a small improvement that can be made. You don’t actually need <React.Fragment> here. React only cares about the key on the top-level element returned from the map. Since the key is already on the <Box> component, the fragment is unnecessary and can be removed entirely.

Spiral-Memory (approved):
LGTM

PR #1114, opened 2026-01-26
Fix blocking busy-wait in typing handler using FIFO async queue

### LINKED ISSUE #1113: Bug: Blocking busy-wait loop in typing handler can freeze the browser
The typing handler in `EmbeddedChatApi` uses a synchronous busy-wait loop to enforce sequential processing of typing events. Because JavaScript is single-threaded, this loop blocks the event loop and can cause the host application to freeze under certain conditions.

---

### Affected File

`packages/api/src/EmbeddedChatApi.ts`

Method: `handleTypingEvent`

---

### Problem

The current implementation relies on a lock enforced via a synchronous loop:

```ts
while (typingHandlerLock) {}
```

Since the event loop is blocked:

* The `setTimeout` callback intended to release the lock cannot execute
* Multiple rapid typing events can lead to a deadlock
* The UI may become unresponsive or freeze entirely

This is especially likely when multiple typing events arrive in quick succession (common in real-time chat environments).

---

### Why This Matters

* Blocks the JavaScript main thread
* Causes degraded UX or complete UI freezes
* Breaks the expected non-blocking behavior of event handling
* High impact in embedded/hosted environments where responsiveness is critical

---

### Expected Behavior

* Typing events should be processed **sequentially**
* The event loop must remain **non-blocking**
* A failure in one typing callback should **not break future typing updates**

---

### Suggested Solution

Replace the synchronous lock with an **asynchronous FIFO queue** (e.g., Promise chaining):

* Guarantees ordering of typing events
* Avoids blocking the event loop
* Preserves existing behavior
* Safer and idiomatic for JavaScript

---

### Notes

This issue is independent of UI concerns and can be addressed entirely within the API layer.
A fix should not change public API behavior, only internal execution semantics.

---

Agarwalchetan:
# Fix blocking busy-wait in typing handler using FIFO async queue

* [x] Removed synchronous busy-wait loop that could block the JavaScript event loop
* [x] Implemented a FIFO async queue using Promise chaining to process typing events sequentially
* [x] Preserved existing typing indicator behavior and event ordering
* [x] Ensured robustness by preventing queue breakage if a typing callback throws
* [x] Kept production logs clean by removing unsafe `console.error` calls
* [x] Added unit tests validating sequential execution and final typing state

Fixes #1113 

---

## Overview

This PR removes a critical performance bug where a synchronous `while` loop was used to lock the thread while handling typing events.
Because JavaScript is single-threaded, this could block the event loop and cause the host application or browser to freeze when multiple typing events occurred in quick succession.

---

## Changes

**File:**
`packages/api/src/EmbeddedChatApi.ts`

**Fix:**

* Replaced `while (typingHandlerLock) {}` with a FIFO async queue implemented via Promise chaining (`this.typingChain`)

**Robustness:**

* Added a `try / catch` block inside the queue consumer to ensure the queue remains functional even if a typing callback throws
* Removed unsafe `console.error` calls to avoid noisy production logs while preserving internal state consistency

**Tests:**

* Added unit tests verifying sequential processing of typing events and correctness of intermediate and final typing states

---

## Video / Screenshots

N/A (internal logic change; no UI modifications)

---

## PR Test Details

* Unit tests added for the typing handler logic
* Manual verification performed by simulating rapid typing events
* No public API changes

---

This change replaces a blocking lock with an async FIFO queue to prevent UI freezes while preserving typing event ordering.

### DIFF
--- a/packages/api/src/EmbeddedChatApi.test.ts
+++ b/packages/api/src/EmbeddedChatApi.test.ts
@@ -1,0 +1,56 @@
+import { describe, test, expect, jest, beforeEach } from '@jest/globals';
+import EmbeddedChatApi from './EmbeddedChatApi';
+
+// Mocks for external dependencies to isolate the unit test
+jest.mock('@rocket.chat/sdk', () => ({
+    Rocketchat: class MockRocketChat {
+        connect() { }
+        resume() { }
+        subscribeRoom() { }
+        onMessage() { }
+        onStreamData() { }
+        subscribeNotifyUser() { }
+        unsubscribeAll() { }
+        disconnect() { }
+    }
+}));
+
+jest.mock('@embeddedchat/auth', () => ({
+    RocketChatAuth: class MockAuth {
+        constructor() { }
+    },
+    IRocketChatAuthOptions: {}
+}));
+
+describe('EmbeddedChatApi Typing Handler', () => {
+    let api: any;
+
+    beforeEach(() => {
+        api = new EmbeddedChatApi('http://localhost:3000', 'GENERAL', {} as any);
+    });
+
+    test('FIFO Async Queue: updates typing status sequentially without blocking', async () => {
+        const statuses: string[][] = [];
+        api.addTypingStatusListener((users: string[]) => {
+            statuses.push([...users]);
+        });
+
+        // Simulate rapid firing of events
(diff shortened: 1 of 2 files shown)


Agarwalchetan:
Hi! Just checking in to see if this PR looks okay or if any changes are needed. Happy to make updates if required.

PR #1126, opened 2026-01-29
fix: cleanup OAuth event listeners and interval on login success

### LINKED ISSUE #1125: Memory leak in OAuth authentication flow
### Description:

Memory leak in OAuth authentication flow where event listeners and interval timers are not cleaned up when login succeeds. The cleanup only happens when the popup is manually closed by the user.

### Steps to reproduce:

1. Enable OAuth authentication (`authFlow: 'OAUTH'`)
2. Click the login button to trigger `loginWithRocketChatOAuth()`
3. Complete OAuth login successfully
4. Check browser's event listeners and active intervals

### Expected behavior:

After successful OAuth login, both the message event listener and the interval timer should be immediately removed from memory.

### Actual behavior:

The event listener on `window` and `setInterval` continue running even after successful login. They only get cleaned up when the interval detects the popup is closed (up to 1 second delay), and if timing is unlucky, they may persist indefinitely. This causes memory leaks when users authenticate multiple times in a session.

Additionally, if the user closes the OAuth popup without completing login, the promise never resolves or rejects, leaving the app in a hanging state.

Anshumancanrock:

- [x] Event listener and interval are cleaned up immediately after successful OAuth login
- [x] Cleanup occurs when user closes popup without completing login
- [x] No race conditions when message and popup close happen simultaneously
- [x] API call errors are properly handled and rejected
- [x] Memory leaks eliminated for repeated OAuth authentication flows

Fixes #1125 

## Description

This PR fixes a memory leak in the OAuth authentication flow where event listeners and interval timers persisted after login completion.

**Before:**

Event listeners and intervals weren't cleaned up on login success, causing memory leaks and hanging promises when popup was closed manually.

**After:**

Added cleanup() function with settled flag to immediately remove listeners/intervals on success or popup close, preventing race conditions and properly handling API errors.

## PR Test Details

**Testing OAuth flow:**
1. Enable OAuth authentication in your EmbeddedChat config
2. Click login button to trigger OAuth flow
3. Complete login successfully → verify no console errors and authentication works
4. Repeat login multiple times → check browser's event listeners don't accumulate
5. Close popup without logging in → verify proper error message is displayed

### DIFF
--- a/packages/auth/src/loginWithRocketChatOAuth.ts
+++ b/packages/auth/src/loginWithRocketChatOAuth.ts
@@ -38,29 +38,47 @@ const loginWithRocketChatOAuth = async (config: { api: Api }) => {
     redirect_uri,
     client_id
   );
-  let params = `scrollbars=no,resizable=no,status=no,location=no,toolbar=no,menubar=no,
-width=800,height=600,left=-1000,top=-1000,rel=opener`;
+  let params = \`scrollbars=no,resizable=no,status=no,location=no,toolbar=no,menubar=no,
+width=800,height=600,left=-1000,top=-1000,rel=opener\`;
   const popup = window.open(authorizeUrl, "Login", params);
 
-  return new Promise<any>((resolve) => {
+  return new Promise<any>((resolve, reject) => {
     if (popup) {
+      let checkInterval: ReturnType<typeof setInterval>;
+      let settled = false;
+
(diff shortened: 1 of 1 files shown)


Anshumancanrock:
hi @Spiral-Memory , Could you please review this PR? Thanks !

PR #1134, opened 2026-01-30
Fix unsafe null destructuring in auth credentials

### LINKED ISSUE #1133: Unsafe null destructuring in auth credentials
### Description:

When users aren't authenticated, API methods in `EmbeddedChatApi.ts` send the literal string `"undefined"` in auth headers instead of empty values. This happens because of unsafe destructuring: `(await this.auth.getCurrentUser()) || {}` returns an empty object when the user is null, causing `userId` and `authToken` to be `undefined`.

### Steps to reproduce:

1. Call any API method when user is not logged in
2. Check the HTTP request headers
3. See `X-Auth-Token: "undefined"` and `X-User-Id: "undefined"` (literal strings)

### Expected behavior:

Auth headers should contain empty strings `""` when user is not authenticated, not the literal string `"undefined"`.

### Actual behavior:

37 API methods send `"undefined"` as header values due to the unsafe pattern `|| {}`. Headers.set() converts JavaScript `undefined` to the string `"undefined"`.

Anshumancanrock:

- [x] Replace unsafe `|| {}` pattern with nullish coalescing `?? { userId: '', authToken: '' }`
- [x] Fix all 37 occurrences across API methods
- [x] Ensure empty strings are sent instead of literal `"undefined"` string in headers
- [x] Maintain backward compatibility with existing UI code

Fixes #1133 

## Description

Fixed a bug where unauthenticated API calls were sending the literal string `"undefined"` in HTTP headers instead of empty values. The original pattern `(await this.auth.getCurrentUser()) || {}` caused destructuring of an empty object, resulting in `undefined` values that JavaScript's `Headers.set()` converted to the string `"undefined"`.

**Changes:**

 `|| {}` → `?? { userId: '', authToken: '' }` (37 occurrences)

**Impact:** Headers now contain clean empty strings instead of `"undefined"`.

### DIFF
--- a/packages/api/src/EmbeddedChatApi.ts
+++ b/packages/api/src/EmbeddedChatApi.ts
@@ -397,7 +397,7 @@ export default class EmbeddedChatApi {
 
   async updateUserNameThroughSuggestion(userid: string) {
     try {
-      const { userId, authToken } = (await this.auth.getCurrentUser()) || {};
+      const { userId, authToken } = (await this.auth.getCurrentUser()) ?? { userId: '', authToken: '' };
       const response = await fetch(
         `${this.host}/api/v1/users.getUsernameSuggestion`,
         {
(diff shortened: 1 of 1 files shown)


Anshumancanrock:
Hi @Spiral-Memory , Could you please review this PR ? Thanks !

PR #1135, opened 2026-02-03
fix: resolve ReferenceError in EmbeddedChat and cleanup API logic

### LINKED ISSUE #1149: Search API does not URL-encode searchText query parameter
### Description:
The search API request does not URL-encode user-provided `searchText` before appending it to query params.

In `packages/api/src/EmbeddedChatApi.ts` (around line 1124), the URL is built with:
`...&searchText=${text}`

Since `text` comes from user input (`packages/react/src/views/MessageAggregators/SearchMessages.js`, line 15), special characters like `&`, `?`, `#`, `%` can break or alter query parsing.

### Steps to reproduce:
1. Open chat and use **Search Messages**.
2. Enter a query containing special characters, e.g. `hello&room?x#tag%`.
3. Trigger search and inspect request/query behavior.
4. Observe incorrect parsing or unexpected search results.

### Expected behavior:
`searchText` should be safely encoded (via `URLSearchParams` or `encodeURIComponent`) so all user input is treated as data, not query syntax.

### Actual behavior:
Raw user input is interpolated directly into the URL query string, which can corrupt query parameters and change request interpretation.

vivekyadav-3:
Hey there! 
I've been working on improving EmbeddedChat's stability and fixing some bugs I ran into while testing. This PR bundles together a few important fixes and improvements that should make the chat experience smoother for everyone.
 What's Fixed

 Search finally works with special characters! (Fixes #1149)

So I was testing the search feature and realized it completely breaks when you search for things like "hello&world" or "test?query". Turns out we weren't encoding special characters (`&`, `?`, `#`, `%`) before sending them to the API, which was causing the URL to get all messed up.

I added proper encoding so now you can search for literally anything without the app freaking out. Go ahead and try searching for `hello&world?test#tag%` - it should just work now!
 No more crashes during login failures

Found a nasty bug where the app would crash if login failed. The notification system was trying to show an error message before it was even ready. Fixed the order of things so errors get displayed properly instead of crashing everything.

Performance Tweaks
 Typing indicator feels snappier now

The "typing..." indicator was hanging around for 15 seconds after someone stopped typing, which felt kinda sluggish. I dialed it down to 10 seconds so it disappears faster and the chat feels more responsive. Small change, but it makes conversations feel more real-time.

 Code Quality Stuff

I also cleaned up a bunch of things under the hood:

Better API handling - Switched from manually building request strings to proper JSON serialization. Less error-prone and handles weird characters way better.

Stronger type checking** - Added validation for UI components so we catch mistakes earlier instead of at runtime.
Cleaner code** - Fixed some dependency warnings, cleaned up hooks, and made the codebase more consistent overall. Also fixed typos in comments and docs.

 Why This Matters

For users:
- Search actually works reliably now, even with special characters
- No more random crashes when login fails
- Chat feels more responsive with the faster typing indicator

For developers:
- Code is more maintainable and catches errors earlier
- Better type safety means fewer bugs slip through
- More consistent patterns across the codebase

 Testing Done

I've tested everything locally and made sure:
-  Project builds without errors
-  Search works with all kinds of special characters
-  Login errors show up properly instead of crashing
-  Typing indicator times out correctly
-  All linting checks pass
-  No breaking changes

Ready for review whenever you get a chance! Let me know if you have any questions or want me to test anything else.


vivekyadav-3:
Closing in favor of a cleaner, more targeted fix introduced in a new PR

PR #1152, opened 2026-02-11
fix: prevent crash when typing unknown slash commands

### LINKED ISSUE #1144: Bug : Crash when typing unknown slash command (e.g., /hello)
This PR fixes a crash that occurs when a user types an unknown slash command (e.g., `/hello`) and presses Enter.

Previously, entering a slash command that does not exist in the client-side commands list caused the application to crash with a `TypeError`. This PR ensures that the client handles unknown slash commands gracefully and aligns behavior with standard Rocket.Chat expectations.

---

## Steps to Reproduce

1. Open EmbeddedChat.
2. Type `/hello` (or any unknown slash command) in the chat input.
3. Press **Enter**.

---

## Observed Behavior

The application crashes with the following error:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'command')
    at CommandsList.js:33
    at HTMLDocument.handleKeyPress (CommandsList.js:54)
```

Image

Image

---

## Expected Behavior

* The message `/hello` should be sent to the server.
* The server should respond with `No such command: /hello` (standard Rocket.Chat behavior).
* At a minimum, the client should not crash.
 * Add a defensive check to ensure `selectedItem` exists before accessing its properties.

vivekyadav-3:
Fixes #1144 – Prevent crash on unknown slash commands

While testing slash commands, I found a bug where entering a command that doesn’t exist (for example /hello) could crash the app with a TypeError.

Root cause

We were accessing properties on a command object without first confirming the command actually exists. When an unknown command was submitted, the code tried to read from undefined, which caused the crash.

Fix

Added a guard to ensure a command is selected before trying to use it:

if (selectedItem) {
  handleCommandClick(selectedItem);
}

Now, if a user types an unknown slash command, the app safely ignores it instead of crashing.

### DIFF
--- a/packages/react/src/hooks/useRCAuth.js
+++ b/packages/react/src/hooks/useRCAuth.js
@@ -25,45 +25,67 @@ export const useRCAuth = () => {
   const handleLogin = async (userOrEmail, password, code) => {
     try {
       const res = await RCInstance.login(userOrEmail, password, code);
+
+      // Handle specific error codes or generic Unauthorized
       if (res.error === 'Unauthorized' || res.error === 403) {
         dispatchToastMessage({
           type: 'error',
-          message:
-            'Invalid username or password. Please check your credentials and try again',
+          message: 'Invalid username or password. Please check your credentials.',
         });
-      } else {
-        if (res.error === 'totp-required') {
-          setPassword(password);
-          setEmailorUser(userOrEmail);
-          setIsLoginModalOpen(false);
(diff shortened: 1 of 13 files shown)


vivekyadav-3:
Hi! I just force-pushed to clean up this branch. It now strictly contains the 1-line null check to fix the crash. Sorry for the noise earlier! Ready for review when you are.

PR #1157, opened 2026-02-16
fix: unknown slash commands crash & incorrect handling

### LINKED ISSUE #1144: Bug : Crash when typing unknown slash command (e.g., /hello)

deepak0x:
This PR fixes a bug where the app crashed if a user typed an unknown slash command (for example, `/hello`) and pressed Enter.

Now, unknown slash commands are treated as command attempts instead of normal chat messages. If the command does not exist on the server, the user will see a "No such command" message.

---
Closes: #1144

Screenshots
Screenshot from 2026-02-17 01-40-40

### 1. packages/react/src/views/CommandList/CommandsList.js

* Added a safety check in `handleKeyPress`.
* Prevents the app from trying to access data when the command list is empty.
* This fixes the crash.

### 2. packages/react/src/hooks/useShowCommands.js

* Updated the logic so the command list is not shown when there are no matching commands.
* Keeps the UI clean and avoids unexpected behavior.

#### handleCommandExecution

* Now tries to execute all slash commands using the API.
* Even if the command is not in the local list, it is sent to the server.
* If the server returns a 400 error (invalid command), a "No such command" toast is shown.

#### sendMessage

* Prevents slash commands from being sent as normal chat messages.
* Ensures they are handled only as commands.

### DIFF
--- a/packages/react/src/hooks/useShowCommands.js
+++ b/packages/react/src/hooks/useShowCommands.js
@@ -9,7 +9,11 @@ const useShowCommands = (commands, setFilteredCommands, setShowCommandList) =>
       const cursor = e.target.selectionStart;
       const tokens = e.target.value.slice(0, cursor).split(/\s+/);
 
-      if (tokens.length === 1 && tokens[0].startsWith('/')) {
+      if (
+        tokens.length === 1 &&
+        tokens[0].startsWith('/') &&
+        getFilteredCommands(tokens[0]).length > 0
+      ) {
         setFilteredCommands(getFilteredCommands(tokens[0]));
         setShowCommandList(true);
       } else {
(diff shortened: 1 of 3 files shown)


deepak0x:
hey @Spiral-Memory 
Pls check this pr if you get time ;)
Thanks

PR #1209, opened 2026-03-16
Fix: File Gallery does not re-render after deleting a file

### LINKED ISSUE #1208: File Gallery does not update after deleting a file — requires page reload
### Description:
When a file is deleted from the File Gallery sidebar, the deletion succeeds on the server but the file remains visible in the list until the page is reloaded. The UI does not re-render to reflect the deletion.

### Steps to reproduce:
- Open a channel and click the Files sidebar/gallery.
- Hover over any file and open its menu.
- Click Delete and confirm in the modal.
- Observe that the deleted file is still shown in the list.
- Redo the navigation again... only then does the file disappear.

### Expected behavior:
After confirming deletion, the file should be immediately removed from the File Gallery list without requiring a reload.

### Actual behavior:
The file persists in the gallery list after successful deletion. The backend deletes the file correctly, but the local file list state owned by [FileGallery](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is never updated, so the stale list continues to render.

DammyCodes-all:
# Brief Title

- [x] Deleting a file from the File Gallery immediately removes it from the list without requiring a page reload.
- [x]  The delete confirmation modal closes correctly after deletion
- [x]  A success toast is shown on successful deletion; an error toast is shown on failure.
- [x]  The deleted message is also removed from the global message store.

Fixes #1208 

### DIFF
--- a/packages/react/src/views/ChatBody/ChatBody.js
+++ b/packages/react/src/views/ChatBody/ChatBody.js
@@ -42,6 +42,12 @@ const ChatBody = ({
   scrollToBottom,
   clearUnreadDividerRef,
 }) => {
+  const getMaxScrollTop = (messageList) =>
+    Math.max(0, messageList.scrollHeight - messageList.clientHeight);
+
+  const isAtBottom = (messageList) =>
+    getMaxScrollTop(messageList) - messageList.scrollTop <= 1;
+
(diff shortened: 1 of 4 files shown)


DammyCodes-all:
Replaced by #1212

PR #1231, opened 2026-03-21
critical fix: Image attachments return 403

### LINKED ISSUE #1229: Bug: Image attachment previews return 403 Forbidden (uploaded images not visible in chat)
### Description:

After a file is successfully uploaded via EmbeddedChat, the image attachment preview in the chat message fails to load with a 403 Forbidden error. The file uploads correctly to the server, but the  tag rendering the preview has no authentication headers, so RC's default "Block Unauthenticated Access" setting blocks the request.

### Steps to reproduce:

1. Log in and navigate to a channel
2. Click the attachment button and upload any image file
3. Observe the chat message after the upload completes

### Expected behavior:

The uploaded image should render as a preview inline in the chat message.
### Actual behavior:

The image preview area is blank/broken. The browser console shows:

Image

Aryan-Verma-999:
fix: Image attachments return 403 added Rc auth tokens to image URLs

- [x] Image attachment previews load correctly
- [x] Auth token is only appended to URLs served from the same RC host
- [x] Graceful fallback if auth fetch fails (.catch() handled, no unhandled rejection)

Fixes #1229 

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-1231 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.

### DIFF
--- a/packages/auth/src/RocketChatAuth.ts
+++ b/packages/auth/src/RocketChatAuth.ts
@@ -180,6 +180,22 @@ class RocketChatAuth {
 
   async save() {
     await this.saveToken(this.currentUser.authToken);
+    try {
+      if (typeof window !== "undefined") {
+        const proxyUrl = "/api/proxy-auth";
+        await fetch(proxyUrl, {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({
+            rc_token: this.currentUser.authToken,
+            rc_uid: this.currentUser.userId,
+            host: this.host,
+          }),
+        }).catch(() => null); // Fail silently if no proxy is configured
+      }
+    } catch (e) {
+      // Ignore proxy errors
+    }
     this.notifyAuthListeners();
   }
 
--- a/packages/react/.storybook/middleware.js
+++ b/packages/react/.storybook/middleware.js
@@ -1,0 +1,80 @@
+// packages/react/.storybook/middleware.js
+const express = require('express');
+
+// Simple in-memory session store for development
+const sessions = {};
+
+module.exports = function expressMiddleware(app) {
+  app.use(express.json());
+
+  // 1. Session Setup Endpoint
+  app.post('/api/proxy-auth', (req, res) => {
+    const { rc_token, rc_uid, host } = req.body;
+    if (!rc_token || !rc_uid || !host) {
+      return res.status(400).json({ error: 'Missing parameters' });
+    }
+
+    // Generate a simple session ID
+    const sessionId = Math.random().toString(36).substring(2, 15);
+    sessions[sessionId] = { rc_token, rc_uid, host };
+
+    // Set HTTP-only cookie
+    res.cookie('ec_session', sessionId, {
+      httpOnly: true,
+      secure: false, // For local Storybook (http://localhost:6006)
+      sameSite: 'lax',
+      path: '/',
+    });
+
+    res.json({ success: true });
+  });
+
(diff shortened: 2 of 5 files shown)


Aryan-Verma-999:
Hey @Spiral-Memory, I pushed the fix for the image preview 403.

While testing, I noticed a separate issue where message history isn’t loading. I’m opening an issue #1232  for that now and looking into a fix.

Spiral-Memory:
I didn't get the reason. Why specifically do we need to wrap it in useAuth? Why is it not working? Can you provide an RCA?

Aryan-Verma-999:
the actual issue is that image attachments are loaded through /file-upload/... as browser asset requests, not through our normal authenticated API calls. Our API requests work because they send X-Auth-Token and X-User-Id, but the image request itself does not include those headers, so Rocket.Chat treats it as unauthenticated and returns 403.

so the fix is really about making media URLs use the current authenticated session. A hook/shared auth helper is just one way to keep that in sync.

Spiral-Memory:
Understood, but the current approach seems risky since it exposes the token in query parameters, and manually verifying the RC host doesn’t feel secure. A better solution might be to use an interceptor pattern, like a service worker or fetch the blob directly in React by including headers and then generate an object URL and use that url to display the place.

Aryan-Verma-999:
got it, updating the approach to fetch the blob with auth headers and generate an object URL instead that avoids token exposure in query params entirely. Will push the updated fix shortly

Aryan-Verma-999:
hey @Spiral-Memory , i need your advice on this 
i tried the blob/object URL approach, but it doesn’t work in the current setup.

issue is that protected images are served from /file-upload/..., and when EmbeddedChat is running cross-origin, fetching those files with auth headers triggers a CORS preflight. That route does not seem to allow this flow, so the request fails before we can even create the blob URL.

so the blob approach does not look viable here.

Aryan-Verma-999:
Hey @Spiral-Memory,
i tried both blob fetch and service worker approach last night but kept bumping into cors preflight.
i went with a BFF proxy instead: after login, auth credentials are stored server-side via /api/proxy-auth and a httpOnly session cookie is set. Image attachments now render as ` `(no tokens in URLs, no CORS issues.)
Included a Storybook middleware as a dev reference implementation. Let me know what you think of the approach.

Spiral-Memory:
Hey, this looks good @Aryan-Verma-999.
Just want to confirm one thing: how can you store credentials on the server side? I mean, does Rocket.Chat allow having some data?

Or are you talking about the API package in EmbeddedChat? (That's not a server; it's part of the EmbeddedChat frontend only.)

Spiral-Memory:
Or you're saying, the developers has to provide a proxy url to store it?

Aryan-Verma-999:
yeah, since EmbeddedChat is just a frontend widget, it can't store secure cookies itself. 
The idea here is that developers using embeddedchat would set up these two simple proxy routes (`/api/proxy-auth` and `/api/proxy-media`) on their own backend (like Node, Next.js, or Django) to securely hold the tokens and serve the images.
i just added `.storybook/middleware.js` as a local dev mock so we can actually test this http-only cookie architecture locally in Storybook without getting blocked by Rocket.Chat's legacy cors rules.
 
we can just expose a `mediaProxyUrl` prop in EmbeddedChat so developers can easily point the widget to their own backend proxy. Does this direction sound good to you?

PR #1261, opened 2026-03-31
fix: handle network errors during login to prevent TypeError crash

### LINKED ISSUE #1210: Bug: Uncaught TypeError in useRCAuth on Network Error during login
When a user attempts to log in but encounters a network error (e.g., `net::ERR_NETWORK_CHANGED` or a DNS failure), the `fetch` call in `EmbeddedChatApi.login` throws an error.

The `catch` block in `packages/api/src/EmbeddedChatApi.ts` currently logs the error but implicitly returns `undefined`:
```typescript
// packages/api/src/EmbeddedChatApi.ts
      if (error instanceof ApiError && error.response?.status === 401) {
        const authErrorRes = await error.response.json();
        return { error: authErrorRes?.error };
      }
      console.error(error); // returns undefined implicitly
```

Subsequently, the react hook `useRCAuth.js` attempts to read `res.error`, leading to an uncaught `TypeError`, completely crashing the login flow and preventing any toast error message from being displayed to the user:
```javascript
// packages/react/src/hooks/useRCAuth.js
      const res = await RCInstance.login(userOrEmail, password, code);
      if (res.error === 'Unauthorized' || res.error === 403) { // TypeError: Cannot read properties of undefined (reading 'error')
```

### Steps To Reproduce
1. Open the EmbeddedChat application
2. Go to the offline state or throttle your network connection to drop it mid-request.
3. Attempt to submit the login form.
4. Observe the unhandled promise rejection and UI crash instead of a graceful toast message.

### Expected Behavior
The login flow should robustly handle network errors returned by `RCInstance.login()`, exiting gracefully and optionally displaying a user-friendly error toast.

mufeezhanif:
## Summary

Fixes #1210 — login crashes with `TypeError: Cannot read properties of undefined (reading 'error')` when a network error occurs during authentication.

## Root Cause

The `login` method in `EmbeddedChatApi.ts` catches errors but only returns a structured response for 401 errors. For all other errors (network failures, DNS errors, etc.), it `console.error()`s and implicitly returns `undefined`. The calling code in `useRCAuth.js` then tries to access `res.error`, which throws a `TypeError`.

### `packages/api/src/EmbeddedChatApi.ts`
- Return `{ error: error.message }` (or `{ error: 'unknown-error' }`) for all non-401 errors instead of implicitly returning `undefined`

### `packages/react/src/hooks/useRCAuth.js`
- Added a null guard on the login response as a defensive check
- Shows a user-friendly toast message ("Unable to connect to server...") if the response is missing

## Testing

- Verified the fix handles the case described in the issue (network error → graceful toast instead of crash)
- The existing catch block in `handleLogin` still handles truly unexpected errors

### DIFF
--- a/packages/api/src/EmbeddedChatApi.ts
+++ b/packages/api/src/EmbeddedChatApi.ts
@@ -143,6 +143,7 @@ export default class EmbeddedChatApi {
         return { error: authErrorRes?.error };
       }
       console.error(error);
+      return { error: error instanceof Error ? error.message : 'unknown-error' };
     }
   }
 
--- a/packages/react/src/hooks/useRCAuth.js
+++ b/packages/react/src/hooks/useRCAuth.js
@@ -25,6 +25,14 @@ export const useRCAuth = () => {
   const handleLogin = async (userOrEmail, password, code) => {
     try {
       const res = await RCInstance.login(userOrEmail, password, code);
+      if (!res) {
+        dispatchToastMessage({
+          type: 'error',
+          message:
+            'Unable to connect to server. Please check your connection and try again.',
+        });
+        return;
+      }
       if (res.error === 'Unauthorized' || res.error === 403) {
         dispatchToastMessage({
           type: 'error',


mufeezhanif:
> [CLA assistant check](https://cla-assistant.io/RocketChat/EmbeddedChat?pullRequest=1261) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/RocketChat/EmbeddedChat?pullRequest=1261) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/RocketChat/EmbeddedChat?pullRequest=1261) it.

I have signed the the agreement!

PR #1272, opened 2026-04-06
fix(ui): prevent message timestamp overlap in small containers

priyapathak39:
Changes
- Removed viewport-based media queries
- Added `flex-wrap: wrap` to the message header
- Updated timestamp styling using:
  - margin-left: auto
  - min-width: fit-content
  - flex-shrink: 0

Why this change?
EmbeddedChat is used inside flexible containers, not just full-screen views. 
Using media queries based on viewport width does not handle all cases.

This solution ensures the layout adapts correctly regardless of container size.

 Result
- No timestamp overlap
- Works in small containers and embedded views
- Maintains clean responsive layout

### DIFF
--- a/packages/react/src/views/Message/Message.styles.js
+++ b/packages/react/src/views/Message/Message.styles.js
@@ -1,3 +1,4 @@
+
 import { css } from '@emotion/react';
 
 export const getMessageStyles = ({ theme }) => {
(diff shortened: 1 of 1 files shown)


priyapathak39:
Hi @vignesh,

Thanks for your feedback!

- Removed viewport-based media queries
- Updated layout to rely on flex-wrap
- Added min-width and flex-shrink constraints to timestamp

Tested in narrow containers and confirmed no overlap.

Please let me know if anything else needs improvement 🙂

priyapathak39:
Hi @vignesh,
I've addressed the feedback from the previous review by:
Removing viewport-based media queries.
Updating the layout to use flex-wrap.
Adding min-width and flex-shrink constraints to the timestamp.

I’ve confirmed that the layout remains stable in narrow containers without overlap.

It's been 3 days,When you have a moment, could you please take another look?

 I’m looking forward to getting this merged.

Thanks!

PR #1273, opened 2026-04-06
fix(auth): resolve context loss causing TypeError in secure auth methods

### LINKED ISSUE #1225: Bug: Secure token functions crash with TypeError — users on secure auth flow cannot log in or log out
### Description

In `packages/react/src/lib/auth.js`, the secure token storage functions (`saveTokenSecure`, `getTokenSecure`, `deleteTokenSecure`) reference `this.handleSecureLogin(...)` — but they are plain standalone `async` functions, not class methods. When invoked via `getTokenStorage(secure: true)`, `this` is `undefined` in strict mode, causing an immediate `TypeError` crash. Any user or deployment configured to use the secure auth flow **cannot log in or log out**.

**File:** `packages/react/src/lib/auth.js` — lines 28–33

### Steps to Reproduce

1. Configure EmbeddedChat with `secure: true` (or any auth flow that calls `getTokenStorage(true)`)
2. Launch the app and attempt to log in
3. Observe a `TypeError: Cannot read properties of undefined (reading 'handleSecureLogin')` in the browser console
4. The user remains stuck on the login screen with no visible error message in the UI

### Expected Behavior

Logging in and out should work correctly when the secure auth flow is configured. The token should be saved and deleted via the secure handler without any runtime errors.

### Actual Behavior

A `TypeError` is thrown as soon as login is attempted on the secure flow:

The crash is thrown inside an async chain, so the UI shows no meaningful error — the user is silently stuck on the login screen or unable to log out.

### Environment

- Affects all environments where `secure: true` is configured
- The `localStorage` (non-secure) auth path is completely unaffected
- No specific OS or browser dependency

### Additional Information

**Root Cause:**
`deleteTokenSecure` (and related functions) use `this.handleSecureLogin(...)`, but are defined as plain `async function` declarations. In strict mode (enforced by all modern bundlers), `this` is `undefined` when these functions are called outside of a class/object context.

**Buggy code:**
```js
async function deleteTokenSecure() {
  this.handleSecureLogin('delete'); // 💥 TypeError in strict mode
}

amankv1234:
## Description
This PR resolves Issue #1225 where users logged in via the secure authentication flow were encountering a `TypeError: Cannot read properties of undefined (reading 'handleSecureLogin')` crash.

**Root Cause:**
In modern environments enforcing ES strict mode, `this` evaluates to `undefined` for standalone function declarations at the module level. When `saveTokenSecure`, `getTokenSecure`, and `deleteTokenSecure` were passed across modules into `RocketChatAuth`, the transpiler stripped their `this` bindings. Consequently, attempting to call `this.handleSecureLogin` within them resulted in an immediate crash.

**Solution:**
Refactored `getTokenStorage` to return the token handlers as object literal methods directly. Moving them from standalone function declarations into object methods ensures lexical scoping is preserved properly when executed within the boundaries of the `RocketChatAuth` instance, thus maintaining the `this` context correctly. 

Fixes #1225

## Changes Made
- Modified `packages/react/src/lib/auth.js` to return object methods directly in `getTokenStorage`.
- Removed top-level standalone function declarations (`saveTokenSecure`, etc.) that were losing context.

## Steps to Test
1. Set up EmbeddedChat with `secure: true` configuration.
2. Launch the application and initiate the login flow.
3. Observe successful authentication without console exceptions.
4. Test the logout flow and ensure tokens are deleted reliably.


amankv1234:
Hi @7vignesh,

I have submitted this pull request with improvements related to type safety and code cleanup.
Could you please review it when you get a chance?

I would really appreciate your feedback, and if everything looks good, kindly consider merging it.

Thank you!

amankv1234:
Hi @7vignesh,

I have updated the PR to address all your feedback.

PR #1277, opened 2026-04-06
fix(api): encode query params in search and user lookup URLs

### LINKED ISSUE #1276: bug: search queries with &, #, or + return wrong results — URL params not encoded
Searching for anything with `&`, `#`, or `+` in it returns wrong results or nothing at all.

`getSearchMessages` in `EmbeddedChatApi.ts` (line 1124) builds the URL with a template literal:

```typescript
`${this.host}/api/v1/chat.search?roomId=${this.rid}&searchText=${text}`
```

The `text` param goes straight into the URL without encoding. So:

- `hello & goodbye` — the `&` starts a new query param. Server receives `searchText=hello ` and a stray `goodbye` param
- `#channel` — the `#` is treated as a URL fragment. `searchText` arrives as an empty string
- `hello world+test` — the `+` is read as a space. Server gets `hello world test`

Same pattern in three other methods:
- `getUserStatus` (line 1236) — `userId=${reqUserId}`
- `userInfo` (line 1253) — `userId=${reqUserId}`
- `userData` (line 1270) — `username=${username}`

The userId ones are less likely to hit this in practice since IDs are usually alphanumeric, but usernames with special characters would break `userData`.

`packages/api/src/EmbeddedChatApi.ts`

deepak0x:
Fixes #1276

`getSearchMessages` builds URLs with template literals and drops user input straight into the query string without encoding. Search for `hello & goodbye` and the server gets `searchText=hello ` because the `&` starts a new param. Search for `#channel` and `searchText` arrives empty because `#` is a URL fragment.

Same thing in `getUserStatus`, `userInfo`, and `userData`. Those are less likely to hit it since user IDs tend to be alphanumeric, but `userData` takes a username and usernames can have special characters.

Added `encodeURIComponent()` to all four.


deepak0x:
hi @Spiral-Memory 
pls review this pr in your free time
Thanks

Issue #1287, opened 2026-04-25
Mention autocomplete triggers on email addresses


deepak0x:
The mention popup in ChatInput fires whenever you type `@` after a space. Problem is, that also matches email addresses — typing something like `contact admin @example.com` pops up the member list and starts filtering against "example.com".

The trigger logic in `useSearchMentionUser.js:23-26`:

```js
if (
  lastChar === '@' &&
  (message.length === 1 || message[message.length - 2] === ' ')
)
```

It checks "is this an `@` after a space?" but doesn't try to figure out whether the user is mentioning someone or just typing an email.

### Steps to reproduce:

1. Open EmbeddedChat in any room
2. In the message input, type `contact admin @example.com`
3. The mention popup appears as soon as you hit `@`
4. It filters members against "example.com"

### Expected behavior:

Mention popup shouldn't fire when the user is typing an email address.

### Actual behavior:

Popup fires on any `@` preceded by a space. You have to dismiss it or ignore it while finishing the email.

### Code reference:

`packages/react/src/hooks/useSearchMentionUser.js:23-26`

deepak0x:
@Spiral-Memory found this while looking at mention-related code paths. The fix would be to add a lookahead check after the `@` trigger — if the characters following `@` contain a dot before any space (like `example.com`), skip triggering the popup. This keeps the existing trigger simple while filtering out email-like patterns. Happy to pick this up.

yash113gadia:
I'd like to work on this!

Looking at `useSearchMentionUser.js:23-26`, a couple of observations on the current behavior:

- A normal inline email like `john@example.com` actually *doesn't* trigger the popup, since the `@` is preceded by a word character rather than a space.
- The popup also already auto-hides when the query matches no members (`setShowMembersList(filteredMentionMembers.length > 0)`).

So the remaining issue is the case in the report: typing `@` after a space followed by a **domain** (e.g. `@example.com`), where the popup flickers and filters against partial matches while you finish typing.

My proposed approach: once in mention-reading mode, suppress the popup when the text after `@` looks like a domain (a `.` followed by TLD-like characters). That covers the email/domain case without affecting normal `@username` mentions.

One edge case I want to confirm before coding: Rocket.Chat usernames can contain dots — do you have a preferred heuristic so we don't accidentally suppress a legitimate mention like `@john.doe`? Happy to go with whatever direction you'd prefer.

Issue #1286, opened 2026-04-25
Member list silently truncated at 100 — "Showing N of N" shows array length instead of actual total


deepak0x:
`getChannelMembers()` in `EmbeddedChatApi.ts:1082` calls `/channels.members?roomId=X` without `count` or `offset` parameters. Rocket.Chat defaults to returning 100 results per page, so rooms with more than 100 members silently lose the rest.

The Members sidebar then displays "Showing 100 of 100" because `RoomMember.js:166` uses `filteredMembers.length` for both numbers — the actual `usersCount` from the `/rooms.info` response (already stored in `channelInfo` via Zustand) is never read.

This also affects the mention autocomplete (`@` popup in ChatInput), which only searches the first 100 members.

### Steps to reproduce:

1. Join or create a room with more than 100 members
2. Open the Members sidebar
3. Observe "Showing 100 of 100" even though the room has more members
4. Try to `@`-mention a user beyond the first 100 — they won't appear in the autocomplete

### Expected behavior:

The sidebar should show "Showing 100 of 350" (using `channelInfo.usersCount` for the total) and ideally paginate or fetch all members.

### Actual behavior:

Shows "Showing 100 of 100". No indication that members are missing. Mention autocomplete can't find users beyond the first page.

deepak0x:
@Spiral-Memory found this while looking at the member list code. The fix would be to pass `count=0` in the `getChannelMembers` API call to fetch all members, and use `channelInfo.usersCount` (already in the Zustand store from the `/rooms.info` call) for the "of N" total in `RoomMember.js:166` instead of reusing the array length for both numbers. Happy to pick this up.

PR #1304, opened 2026-05-22
fix: resolve email mention trigger and paginate member list retrieval

vivekyadav-3:
Resolves both **Issue #1287** and **Issue #1286**.

#### 1. Fix Issue #1287: Mention autocomplete triggers on email addresses
- Prevented the mention suggestions popup from triggering or remaining visible when the search query portion after `@` contains a dot (`.`).
- This filters out domain/email-like typing patterns (e.g. `admin@example.com`) dynamically and self-corrects if the dot is deleted.
- File modified: `packages/react/src/hooks/useSearchMentionUser.js`

#### 2. Fix Issue #1286: Member list silently truncated at 100
- Implemented client-side pagination in the API client (`getChannelMembers`) to fetch members in batches of 100 using `offset` and `count` parameters until all members are fetched.
- Updated the sidebar to display the actual count of members using the `channelInfo.usersCount` value from Zustand store (e.g., `Showing 100 of 350` instead of `Showing 100 of 100`).
- Files modified: `packages/api/src/EmbeddedChatApi.ts`, `packages/react/src/views/RoomMembers/RoomMember.js`

### DIFF
--- a/packages/api/src/EmbeddedChatApi.ts
+++ b/packages/api/src/EmbeddedChatApi.ts
@@ -1092,20 +1092,44 @@ export default class EmbeddedChatApi {
     const roomType = isChannelPrivate ? "groups" : "channels";
     try {
       const { userId, authToken } = (await this.auth.getCurrentUser()) || {};
-      const response = await fetch(
-        `${this.host}/api/v1/${roomType}.members?roomId=${this.rid}`,
-        {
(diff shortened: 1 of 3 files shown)


vivekyadav-3:
cc: @sidmohanty11 @Spiral-Memory

PR #1334, opened 2026-06-25
feat/ai-adapter

Aryan-Verma-999:
# Pluggable AI Adapter Layer
Adds a zero-dependency @embeddedchat/ai-adapter package that lets host applications inject any LLM provider (Ollama, OpenAI, Gemini, or custom) into EmbeddedChat without adding AI dependencies to the core library.

- [x] New @embeddedchat/ai-adapter package with IAIAdapter interface, BaseAIAdapter base class, and adapter implementations for Ollama, OpenAI, and Gemini
- [x] AI is entirely opt-in, passing no aiAdapter prop leaves the UI identical to today, zero footprint
- [x] ✨ Reply suggestions - click to get 3 AI-generated reply chips above the input
- [x] 📝 Chat summary - modal with a concise summary of the conversation

## Video/Screenshots

https://github.com/user-attachments/assets/0070c236-322d-496c-8261-9d18617f3268

## PR Test Details

**Note**: The PR will be ready for live testing at https://rocketchat.github.io/EmbeddedChat/pulls/pr-1334 after approval. Contributors are requested to replace `<pr_number>` with the actual PR number.


Aryan-Verma-999:
Follow up with a few new commits:

**AI Message Composer** - AI-assisted message drafting directly from the composer
**Thread Summarizer** - Generate concise summaries for individual threads
**AI Theme** - Added a dedicated AI-themed UI for AI-powered features

Also attaching a short demo showcasing the new additions.

https://github.com/user-attachments/assets/a92c9619-4af1-4af9-95c9-596e89320f04

Spiral-Memory:
Hi @Aryan-Verma-999,

I went through the videos. You're moving in the right direction, but it's still not fully aligned with what we discussed.

Here are the improvements I'd like to see as early as possible:

**1. Summarise Chat / Suggest Replies**

* The icons should be very subtle. The current design looks dated and doesn't feel suitable for the product.
* Showing the summary in a popup is not an optimal experience.
* The summary should appear as a **"You only"** message, meaning only the person who generated it can see it similar to how user messages looks like. Use an AI bot icon or something similar. for avatar The design is very important here, so keep it clean, subtle, and polished.
* Instead of **"Summarise"**, use a title like **"Catch up"** so it's clear that only recent messages are being summarised.
* You can take inspiration from how Google Chat handles this with Gemini. It's implemented very well.
* Suggested replies should be generated automatically. There shouldn't be a separate button for them.
* For chat summarisation, keep the **Catch up** action at the top near the maximise icon so the experience is consistent for both channels and threads.

**2. Message Composer**

* Showing a long list of options when text is selected is not the expected experience.
* We've already discussed that the message composer should follow the same standards as the one Ayush built for Rocket.Chat last year.
* Use a similar UI approach with subtle actions, inline text replacements, and a clean experience. It should not use an AI suggestion notification bar.
* Please make sure you're following clean UI standards. What you've built is like a POC, but it's not yet at a product-ready quality level.

Please update the implementation based on the above feedback and let me know once it's genuinely ready for another review.

Aryan-Verma-999:
Thanks for the detailed feedback, @Spiral-Memory. I’ve reworked the AI experience to focus on a cleaner, product-ready UI:

- Added a subtle header-level Catch up action with private “You only” summary messages.

- Removed duplicate thread actions and the summary popups.
- Reply suggestions now generate automatically and use the latest conversation context.
- Rebuilt the composer around inline, streamed AI changes with accept/reject controls.
- Reduced actions to Fix grammar, Shorten, Translate, and Emojify.
- Tightened AI prompts for more reliable output.


It’s ready for another review.
here is the demo video for the same


https://github.com/user-attachments/assets/3d844c36-4ce2-402c-99dc-463c0f10277a

Spiral-Memory:
This iteration looks much better. I'd like to test it with a longer sentence by highlighting a section, applying the AI enhancement, and checking how it handles the changes. Our main goal is to make sure edits across multiple lines work properly. If this is not addressed yet, I'm okay approving the PR, but I'd appreciate seeing the results soon.

packages/ai-adapter/src/adapters/MockAdapter.ts
```diff
@@ -0,0 +1,18 @@
+// For testing/demo only — returns hardcoded responses, requires no API key
```
Spiral-Memory: You can either remove MockAdapter or include it as part of your test suite.

packages/ai-adapter/src/adapters/GeminiAdapter.ts
```diff
@@ -0,0 +1,126 @@
...
+        "Content-Type": "application/json",
+        ...this.config.headers,
+      },
+      body: JSON.stringify({
+        contents,
+        systemInstruction: {
+          parts: [
+            {
+              text: context.metadata?.composerTransformation
+                ? "You perform exact composer transformations. Return only the requested transformed source text, with no explanation or chat reply."
```
Spiral-Memory: Can we not hardcode system prompts ?
Why we will have system promots in adapter files ?

packages/ai-adapter/README.md
```diff
@@ -0,0 +1,149 @@
...
+const adapter = new OpenAIAdapter({ apiKey: process.env.OPENAI_API_KEY });
+
+<EmbeddedChat
+  host="https://chat.example.com"
+  roomId="GENERAL"
+  aiAdapter={adapter}
+/>
+```
+
+When `aiAdapter` is provided, a ✨ button appears in the message input toolbar. Clicking it calls `getSuggestions()` with the recent conversation history and displays clickable reply chips above the input.
```
Spiral-Memory: Please update your README

packages/ai-adapter/README.md
```diff
@@ -0,0 +1,149 @@
...
+When `aiAdapter` is **not** provided: zero UI changes, zero bundle size impact.
+
+## Built-in Adapters
+
+### OpenAIAdapter
+
+```typescript
+import { OpenAIAdapter } from '@embeddedchat/ai-adapter';
+
+const adapter = new OpenAIAdapter({
```
Spiral-Memory, line 69: Configure the system prompt so users can select the appropriate adapter for each task type. Additionally, a single adapter can support multiple models, each with its own customized system prompt.

packages/react/src/theme/AITheme.js
```diff
@@ -0,0 +1,102 @@
+const AITheme = {
```
Spiral-Memory: No need for this theme

Spiral-Memory (changes requested):
Everything else looks good; please address the comments. Thanks.

packages/react/src/theme/AITheme.js
Aryan-Verma-999: removed

packages/ai-adapter/README.md
Aryan-Verma-999: done

packages/ai-adapter/src/adapters/GeminiAdapter.ts
Aryan-Verma-999: Addressed. I removed the hardcoded prompts from the provider adapters.

packages/ai-adapter/src/adapters/MockAdapter.ts
Aryan-Verma-999: removed

Aryan-Verma-999:
Added a video demonstrating AI composer edits across multiple selected lines. The selected section is replaced inline, preserves the surrounding message content, and can handle separate AI changes within the same draft.

https://github.com/user-attachments/assets/184e544d-ad3e-4fc6-89b6-4949f7c98be1

Spiral-Memory (approved):

Spiral-Memory:
Looks good to me. We'll plan for more addition and optimization, but for the initial version. This looks great. Thanks for the good work.

